PR #1, opened 2022-12-26
Feature/login screen

vmkhitaryanscn:


src/screens/Login/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+    };
+    name: string;
+    logo?: string;
+  };
+  mobileDataAllowed: boolean;
+  fcmToken: string | null;
+  appLanguage: 'fr' | 'en';
+};
+
+const LoginScreen: FC = () => {
```
BamMironov, line 33: Please rename the filename Login => LoginScreen

src/screens/Login/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+      name: 'MindLogger',
+    },
+  };
+  const { skin } = mockProps;
+
+  const { navigate } = useNavigation();
+  const { t, i18n } = useTranslation();
+  const title = skin.name;
+
+  const changeLanguageHandler = () => {
```
BamMironov, line 49: changeLanguageHandler => setEnglishLanguage

src/features/login/ui/LoginForm.styles.ts
```diff
@@ -0,0 +1,29 @@
...
+    minWidth: 140,
+  },
+  buttonText: {
+    fontSize: 20,
+  },
+  text: {
+    color: colors.secondary,
+    fontSize: 20,
+  },
+});
```
BamMironov, line 29: I don't think we ever need this file

src/screens/index.tsx
```diff
@@ -0,0 +1,19 @@
+import { createNativeStackNavigator } from '@react-navigation/native-stack';
+
+import LoginScreen from './Login';
+
+const Stack = createNativeStackNavigator();
+
+export const Routing = () => {
```
BamMironov, line 7: Routing => Screens

BamMironov (changes requested):

src/shared/ui/StatusBar/index.tsx
```diff
@@ -0,0 +1,10 @@
+import React, { FC } from 'react';
+import { StatusBar as BaseStatusBar, Platform } from 'react-native';
+
+const isIOS = Platform.OS === 'ios';
```
BamMironov, line 4: Please move the constant to /shared/lib/constants.ts and name it IS_IOS. You can also create IS_ANDROID accordingly.

src/app/providers/native-base-provider/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+          },
+        },
+      },
+    },
+    colors: {
+      primary: {
+        50: '#0067A0',
+      },
+      light: {
+        50: '#ffffff99',
```
BamMironov, line 48: image

We have a lot of colors in use. Please declare them all here. 
Let's also set "500" as a default or main color in a palette, and if there are different shades of a color we will set a number according to NativeBase guides: https://docs.nativebase.io/default-theme, the darker a color is, the greater the number.

BamMironov (changes requested):

BamMironov (approved):

PR #7, opened 2023-01-06
M2-339: Sign-Up implementation: UI, Model layers with mocks

anq83:
https://mindlogger.atlassian.net/browse/ML-339

### DIFF
--- a/src/screens/SignUpScreen/index.tsx
+++ b/src/screens/SignUpScreen/index.tsx
@@ -1,5 +1,5 @@
 import { FC } from 'react';
-import { SafeAreaView } from 'react-native';
+import { Keyboard, SafeAreaView, TouchableWithoutFeedback } from 'react-native';
 
 import { SignUpForm } from '@features/sign-up';
 import {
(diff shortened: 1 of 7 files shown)


BamMironov (approved):

src/screens/SignUpScreen/index.tsx
```diff
@@ -13,17 +13,19 @@ import {
 const SignUpScreen: FC = () => {
   return (
     <KeyboardAvoidingView>
-      <SafeAreaView>
-        <StatusBar />
+      <TouchableWithoutFeedback onPress={Keyboard.dismiss}>
```
vmkhitaryanscn, line 16: @BamMironov @anq83 
I suggest we create a Container component, which will include StatusBar, SafeAreaView, and TouchableWithoutFeedback, KeyboardAvoidingView, etc(all configurable with boolean values if needed (KeyboardAvoidingView for example can be true or false)), also we can put global App margins, backgrounds, etc.

vmkhitaryanscn (approved):

PR #13, opened 2023-01-11
M2-361 Implement What Is MindLogger screen

anq83:


src/screens/About/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+    fontWeight: '400',
+    lineHeight: 28,
+  },
+  bullet_list: {
+    marginBottom: 20,
+    color: colors.tertiary,
+  },
+});
+
+export default AboutScreen;
```
BamMironov, line 69: Please rename the file/directory to "AboutScreen.tsx"

src/screens/About/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+import MarkdownView from '@app/shared/ui/MarkdownView';
+import { Box, Text, ScrollView, Image } from '@shared/ui';
+
+const AboutScreen: FC = () => {
+  const { t } = useTranslation();
+
+  return (
+    <Box bg="$secondary">
+      <ScrollView h={'100%'} px={8}>
+        <MarkdownView
```
BamMironov, line 17: We will probably need this component to be reused in the future. Could you please move it with markdownStyle to "shared/ui"?

src/screens/index.tsx
```diff
@@ -8,6 +8,8 @@ import { colors } from '@app/shared/lib';
 import { RootStackParamList } from '@shared/lib/navigation';
 import { Text, CloseIcon } from '@shared/ui';
 
+import packageJson from '../../package.json';
```
BamMironov, line 11: Please create a constant value in shared/lib/constants.tsx that will keep package.json's version and use it here

BamMironov (changes requested):

src/screens/About/index.tsx
anq83: MarkdownView is already in shared. But styles should be passed as prop, because it's specific for every view.

src/screens/index.tsx
anq83: ok

src/screens/About/index.tsx
anq83: ok

BamMironov (approved):

PR #15, opened 2023-01-12
Feature/change language

vmkhitaryanscn:


src/features/change-language/ui/ChangeLanguageSelector.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const ChangeLanguageSelector: FC<BoxProps> = props => {
+  const { t, i18n } = useTranslation();
+  const navigation = useNavigation();
+  const languagesAvailable = Object.keys(i18n.services.resourceStore.data);
+  const { resolvedLanguage } = i18n;
+
+  const changeLanguage = async (locale: string) => {
+    await i18n.changeLanguage(locale);
+    navigation.goBack();
```
BamMironov, line 24: Navigation should be handled on a `screen`'s level

src/features/change-language/ui/ChangeLanguageSelector.tsx
```diff
@@ -0,0 +1,52 @@
...
+  ArrowRightIcon,
+} from '@shared/ui';
+
+const ChangeLanguageSelector: FC<BoxProps> = props => {
+  const { t, i18n } = useTranslation();
+  const navigation = useNavigation();
+  const languagesAvailable = Object.keys(i18n.services.resourceStore.data);
+  const { resolvedLanguage } = i18n;
+
+  const changeLanguage = async (locale: string) => {
```
BamMironov, line 22: As this function is a part of business logic and doesn't depend on react render then please move it to "model" segment.
The function here should look like this:
```
const changeLanguage = (locale: string) => {
  changeLanguageModel.changeLanguage(locale);
  onChange();
}
```

src/features/change-language/ui/ChangeLanguageSelector.tsx
```diff
@@ -0,0 +1,52 @@
...
+    await i18n.changeLanguage(locale);
+    navigation.goBack();
+  };
+
+  return (
+    <Box {...props}>
```
BamMironov, line 28: This is not necessary to rewrite. Just a note.
If you have a `YStack`/`XStack`/`ZStack` as root element in a component then it is not necessary to additionally wrap it in `Box`, because `YStack`/`XStack`/`ZStack` are `Box`es under the hood and accept `BoxProps`.

src/shared/lib/navigation/types.ts
```diff
@@ -1,32 +1,32 @@
+import type { BottomTabScreenProps } from '@react-navigation/bottom-tabs';
 import type { CompositeScreenProps } from '@react-navigation/native';
 import type { NativeStackScreenProps } from '@react-navigation/native-stack';
-import type { BottomTabScreenProps } from '@react-navigation/bottom-tabs';
 
 export type RootStackParamList = {
   Login: undefined;
   SignUp: undefined;
   AboutApp: undefined;
+  ChangeLanguage: undefined;
```
BamMironov, line 9: Well, it was supposed to be the AppLanguage screen. But if you think it should be called ChangeLanguage - I have no problems with that, just please remove AppLanguage

BamMironov (changes requested):

BamMironov (approved):

PR #19, opened 2023-01-16
Feature/profile screen

vmkhitaryanscn:


src/features/logout/index.ts
```diff
@@ -0,0 +1,3 @@
+import LogoutRowButton from './ui/LogoutRowButton';
```
BamMironov, line 1: Let's instead write`export * from './ui'` for the index.ts files. That will make it easier to add further changes

src/screens/SettingsScreen.tsx
```diff
@@ -0,0 +1,49 @@
...
+const SettingsScreen: FC = () => {
+  const { navigate } = useNavigation();
+  const { t } = useTranslation();
+
+  const navigateToAppLanguage = () => {
+    navigate('ChangeLanguage');
+  };
+
+  return (
+    <TouchableWithoutFeedback onPress={Keyboard.dismiss}>
```
BamMironov, line 20: I do not think we need it on this screen component

BamMironov (changes requested):

BamMironov (approved):

PR #21, opened 2023-01-19
Feat/383 411 forgot password, add redux

anq83:
https://mindlogger.atlassian.net/browse/M2-383
https://mindlogger.atlassian.net/browse/M2-411


src/entities/identity/model/actions.ts
```diff
@@ -1,3 +1,3 @@
-import { testIdentitySlice } from './slice';
+import { identitySlice } from './slice';
 
-export const onAuthSuccess = testIdentitySlice.setAuth;
+export const { onAuthSuccess } = identitySlice.actions;
```
BamMironov, line 3: I think we can export actions like this:
`export const actions = identitySlice.actions;`
`export const reducer = identitySlice.reducer;`
or even
`export const { actions, reducer } = identitySlice`;

src/shared/lib/redux/types.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { reduxStore } from '@app/app/providers';
+
+export type RootState = ReturnType<typeof reduxStore.getState>;
+
+export type AppDispatch = typeof reduxStore.dispatch;
+
+export type ThunkAppDispatch = ThunkDispatch<RootState, void, Action>;
+
+export const useThunkDispatch = () => useDispatch<ThunkAppDispatch>;
```
BamMironov, line 12: I would rename it to useAppDispatch because useThunkDispatch implies using thunks but we don't use thunks here.

src/shared/lib/redux/types.ts
```diff
@@ -0,0 +1,12 @@
+import { Action, ThunkDispatch } from '@reduxjs/toolkit';
+import { useDispatch } from 'react-redux';
+
+import { reduxStore } from '@app/app/providers';
```
BamMironov, line 4: This import is prohibited since "app" is the top-level layer. I'd suggest you move these types to ReduxProvider.
You can also declare there RootState and AppDispatch as global types

src/shared/lib/redux/types.ts
BamMironov: Also, it is a good practice to create useAppSelector hook.
Both of them you can move to shared/lib/hooks

src/app/providers/index.tsx
```diff
@@ -3,17 +3,22 @@ import { FC, Suspense, PropsWithChildren } from 'react';
 import { NavigationContainer } from '@react-navigation/native';
 
 import ReactQueryProvider from './ReactQueryProvider';
+import ReduxProvider from './ReduxProvider';
 import TamaguiProvider from './TamaguiProvider';
 
+export { reduxStore } from './ReduxProvider';
```
BamMironov, line 9: Please remove the unused import variable

src/entities/identity/model/hooks/useRegistrationMutation.ts
```diff
@@ -19,13 +21,17 @@ type UseRegistrationReturn = {
 export const useRegistrationMutation = (
   onSuccess: () => void,
 ): UseRegistrationReturn => {
+  const dispatch = useDispatch();
```
BamMironov, line 24: useAppDispatch should be used here

src/screens/AppletsScreen/index.tsx
```diff
@@ -1,9 +1,16 @@
...
+import { useSelector } from 'react-redux';
+
+import { IdentityModel } from '@app/entities/identity';
 import { colors } from '@app/shared/lib';
 import { Text, YStack } from '@shared/ui';
 
 const AppletsScreen: FC = () => {
+  const accessToken = useSelector(IdentityModel.selectors.accessTokenSelector);
+
+  console.log('accessToken', accessToken);
```
BamMironov, line 12: Please clean up `console.log`s

src/shared/api/types.ts
```diff
@@ -1,30 +1,30 @@
...
+// }
+
+// export interface IErrorResponse {
+//   data: IErrorResponseData;
+// }
+
+// export interface IApiError {
+//   response?: IErrorResponse | undefined | null;
+//   message: string;
+// }
```
BamMironov, line 27: Can we remove them all for now?

BamMironov (changes requested):

src/shared/lib/hooks/redux.ts
```diff
@@ -0,0 +1,5 @@
+import { useDispatch, useSelector } from 'react-redux';
+
+export const useAppDispatch = () => useDispatch<ThunkAppDispatch>;
```
BamMironov, line 3: We should use `AppDispatch` type here. If you intend to add a thunks middleware (or whatever middleware) you can do that and if do that the right way, types from middlewares will be correctly inferred to dispatch function.

https://redux-toolkit.js.org/usage/usage-with-typescript#correct-typings-for-the-dispatch-type

BamMironov (changes requested):

BamMironov (approved):

PR #23, opened 2023-01-23
Feat/m2 425

BamMironov:


src/entities/identity/model/selectors.ts
```diff
@@ -1,2 +1,14 @@
...
-  state.identity.accessToken;
+import { createSelector } from '@reduxjs/toolkit';
+
+import { TokenStorage } from '@shared/lib';
+
+const selectIdentity = (state: RootState) => state.identity;
+
+export const selectUser = createSelector(
+  selectIdentity,
+  identity => identity.user,
```
anq83, line 9: I'd prefer to avoid unnecessary linking on selector level.
Can we use ... => state.identity.user instead? Code would be more obvious and understandable

src/jobs/response-interception.ts
```diff
@@ -0,0 +1,43 @@
...
+      const config = error.config as RequestConfig;
+
+      if (error.response?.status === 401 && !config?.retry) {
+        config.retry = true;
+
+        const refreshToken = TokenStorage.getString('refreshToken');
+        const tokenType = TokenStorage.getString('tokenType');
+
+        if (!refreshToken || !tokenType) {
+          return Promise.reject(error);
```
anq83, line 23: Shouldn't we reset tokens and redux/user here to force re-select isAuthenticated ?

src/jobs/response-interception.ts
```diff
@@ -0,0 +1,43 @@
...
+
+        try {
+          const { data } = await IdentityService.refreshToken({ refreshToken });
+
+          TokenStorage.set('accessToken', data.result.accessToken);
+          TokenStorage.set('refreshToken', data.result.refreshToken);
+
+          config.headers.Authorization = `${tokenType} ${data.result.accessToken}`;
+        } catch (e) {
+          Promise.reject(e);
```
anq83, line 34: Shouldn't we reset tokens and redux/user here to force re-select isAuthenticated ?

anq83 (changes requested):

anq83 (approved):

PR #24, opened 2023-01-24
M2-416: Add Applet list and all related components: UI, model layers

anq83:
https://mindlogger.atlassian.net/browse/M2-416


src/screens/AppletsScreen/index.tsx
```diff
@@ -1,24 +1,32 @@
...
-      <Text color={colors.tertiary} fontSize={18}>
-        User is logged in
-      </Text>
-
-      <Text color={colors.tertiary} fontSize={18}>
-        This is mock for applet list
-      </Text>
-    </YStack>
+    <Box bg="$secondary" flex={1}>
+      <ImageBackground
```
BamMironov, line 10: I assume this background image will be used on a lot of screens. Perhaps, that the default image for all screens, except the authentication ones.
I'd suggest using this approach: https://stackoverflow.com/a/70117826

Let's create a BackgroundImage component that wraps the children in the `ImageBackground`. Then we'll wrap the NavigationContainer into BackgroundImage.

src/shared/ui/AppletRoundLogo.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const getImageUrl = (url: string): string => {
+  if (url.endsWith('.svg')) {
+    // todo - remove if unnecessary
+    return url.replace('.svg', '.jpg');
+  }
+  return url;
+};
+
+const AppletRoundLogo: FC<Props> = ({ applet, size = 64 }) => {
```
BamMironov, line 34: Please rename to RoundLogo

src/shared/ui/AppletRoundLogo.tsx
```diff
@@ -0,0 +1,74 @@
...
+declare module 'react-native-svg' {
+  interface LinearGradientProps {
+    children: JSX.Element | JSX.Element[];
+  }
+}
+
+type Props = {
+  applet: Applet;
+  size?: number;
+};
```
BamMironov, line 24: As far as I understand, the types that are actually in use (from the whole `Applet` type): "image", "name".
Let's abstract from "Applet" type here and extend Props like this: 
```
type Props = {
  imageURI: string;
  name: string;
  size?: number;
}
``

BamMironov (changes requested):

BamMironov (approved):

PR #26, opened 2023-01-26
feat(Splash): Added splash screen

BamMironov:


src/shared/lib/contexts/SplashContext.ts
```diff
@@ -0,0 +1,15 @@
+import { createContext, useContext } from 'react';
+
+type SlashContext = {
```
anq83, line 3: Splash io Slash, typo ?

anq83 (changes requested):

src/shared/lib/contexts/SplashContext.ts
BamMironov: Apparently, yes

anq83 (approved):

PR #44, opened 2023-02-02
Feat/m2 506 default screen

BamMironov:
- Added logic responsible for rendering default route.
- Added logic responsible for persisting particular routes.
- Split the `@screens` layer into segments for better maintenance.


src/screens/lib/navigationService.ts
```diff
@@ -0,0 +1,31 @@
+import { createSecureStorage } from '@shared/lib';
+
+import { InitialRoute } from './types';
+
+const storage = createSecureStorage('navigation-storage');
+
+function SessionService() {
```
anq83, line 7: Shouldn't it be named the same as file name?

src/screens/lib/navigationService.ts
```diff
@@ -0,0 +1,31 @@
...
+  return {
+    setInitialRoute(route: InitialRoute) {
```
anq83, line 9: Can we rename it to setStartupRoute?

src/screens/lib/types.ts
```diff
@@ -0,0 +1,6 @@
+import { ScreenParams, ScreenRoute } from '../config';
+
+export type InitialRoute = {
```
anq83, line 3: Maybe StartupRoute?

src/screens/lib/navigationService.ts
```diff
@@ -0,0 +1,31 @@
...
+      storage.set('initialRoute', JSON.stringify(route));
+    },
+
+    clearInitialRoute() {
+      storage.delete('initialRoute');
+    },
+
+    getInitialRoute() {
```
anq83, line 17: the same

src/screens/ui/RootNavigator.tsx
```diff
@@ -0,0 +1,109 @@
...
+            color: colors.white,
+          },
+        }}
+        component={AboutScreen}
+      />
+
+      <Stack.Screen
+        name="Applets"
+        options={{
+          title: 'Applets',
```
anq83, line 86: The Title should be "Hi [User Name]"

anq83 (changes requested):

src/screens/lib/navigationService.ts
BamMironov: Correct

src/screens/lib/navigationService.ts
BamMironov: Not exactly. "Initial route" has been coined by react-navigation so I just follow their API naming

src/screens/lib/types.ts
BamMironov: Please see above

src/screens/lib/navigationService.ts
BamMironov: Please see above

src/screens/ui/RootNavigator.tsx
BamMironov: I can do but I don't think it is related to my PR because it was not me who worked on this screen

anq83 (approved):

PR #45, opened 2023-02-03
M2-519 added change password api implementation, refactor

vmkhitaryanscn:


src/screens/index.tsx
```diff
@@ -8,12 +8,14 @@ import {
 import { useTranslation } from 'react-i18next';
 
 import { APP_VERSION, colors, ENV } from '@app/shared/lib';
+import SettingsScreen from '@screens/SettingsScreen';
```
BamMironov, line 11: Please use relative import here

BamMironov (changes requested):

PR #50, opened 2023-02-07
Feat/m2 526 back button

BamMironov:


src/entities/identity/lib/nameUtils.ts
```diff
@@ -0,0 +1,10 @@
+export function getFirstName(fullName: string): string {
```
anq83, line 1: The function itself looks like workaround, BE should return splitted data

src/screens/ui/RootNavigator.tsx
```diff
@@ -89,21 +97,15 @@ export default () => {
...
         options={{
-          title: 'Applets',
           headerStyle: {
             backgroundColor: colors.lighterGrey2,
           },
           headerTitleStyle: {
             fontSize: 16,
             color: colors.tertiary,
           },
-          headerLeft: () => (
```
anq83, line 100: Can we leave it temporarilly in order to change user if we don't have logout yet?
Or just to have possibility to go to login screen

anq83 (approved):

src/entities/identity/lib/nameUtils.ts
BamMironov: That's exactly what it is. Yes, the back-end should return first and last names as independent fields

src/screens/ui/RootNavigator.tsx
BamMironov: We are implementing logout feature today, so no worries

PR #58, opened 2023-02-13
M2-538 added activity item text component

vmkhitaryanscn:


src/shared/ui/survey/SimpleTextInput.tsx
```diff
@@ -0,0 +1,55 @@
+import { FC, useState } from 'react';
+import { TextInputProps } from 'react-native';
+
+import { useTranslation } from 'react-i18next';
+
+import { colors } from '@shared/lib';
+import { Input } from '@shared/ui';
+
+type ValueType = '' | 'http://www.w3.org/2001/XMLSchema#integer';
```
BamMironov, line 9: I may misinterpret the need for it but I think we can use the following type:
`type KeyboardType = 'text' | 'numeric'`;

src/shared/ui/survey/SimpleTextInput.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+type Config = {
+  isOptionalText: boolean;
+  maxLength: number;
+  removeBackOption: boolean;
+  valueType: ValueType;
```
BamMironov, line 15: Here you can use `keyboardType: KeyboardType;`

BamMironov: or just delete it because it is already a part of `TextInputProps `

src/shared/ui/survey/SimpleTextInput.tsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+type Props = {
+  valueType?: ValueType;
+  config: Config;
```
BamMironov, line 20: Please put all props on the same level. No need to put them in the `config` object in this case

src/shared/ui/survey/SimpleTextInput.tsx
```diff
@@ -0,0 +1,55 @@
...
+} & TextInputProps;
+
+const SimpleTextInput: FC<Props> = ({
+  value = '',
+  onChange,
+  valueType = '',
+  config,
+}) => {
+  const { t } = useTranslation();
+  const [inputValue, setInputValue] = useState(value);
```
BamMironov, line 31: Please rename. value, setValue

src/shared/ui/survey/SimpleTextInput.tsx
```diff
@@ -0,0 +1,55 @@
...
+};
+
+type Props = {
+  valueType?: ValueType;
+  config: Config;
+  onChange: (text: string) => void;
+} & TextInputProps;
+
+const SimpleTextInput: FC<Props> = ({
+  value = '',
```
BamMironov, line 25: Please rename to initialValue

BamMironov (changes requested):

src/shared/ui/survey/SimpleTextInput.tsx
vmkhitaryanscn: We get exactly "valueType" prop from backend, and then I "convert" it to keyboardType.

src/shared/ui/Input.tsx
```diff
@@ -38,6 +40,19 @@ export const InputFrame = styled(
           placeholderTextColor: '$lighterGrey',
           borderBottomColor: '#1e1d1d',
         },
+        survey: {
+          borderBottomWidth: 1,
+          borderBottomColor: '$mediumGrey',
+          width: '100%',
```
anq83, line 46: Layout styles should go upper. See css properties ordering recomendations.

src/shared/ui/survey/SimpleTextInput.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import { useTranslation } from 'react-i18next';
+
+import { colors } from '@shared/lib';
+import { Input } from '@shared/ui';
+
+type Props = {
+  config: {
+    maxLength: number;
+    valueType: '' | 'http://www.w3.org/2001/XMLSchema#integer';
```
anq83, line 12: Why should we relay on schema definitions in Mobile app?
The type might be described with simple Enum value

anq83 (changes requested):

src/shared/ui/survey/SimpleTextInput.tsx
anq83: Then backend should be changed

BamMironov: Agreed. @vmkhitaryanscn,  please contact @Tom1204 in order to change the data contract

src/shared/ui/Input.tsx
vmkhitaryanscn: Do you mean the following? https://tamagui.dev/docs/intro/props#order-is-important

vmkhitaryanscn: @anq83 Do you mean the following? https://tamagui.dev/docs/intro/props#order-is-important

BamMironov (approved):

anq83 (approved):

PR #65, opened 2023-02-21
Feat/m2 243 applets api integration

BamMironov:


src/widgets/activity-group/model/hooks/useActivityGroups.ts
```diff
@@ -19,14 +21,30 @@ type UseActivityGroupsReturn = {
...
-
-  const builder = createActivityGroupsBuilder({
-    allAppletActivities: allActivityMocks,
-    appletId: 'apid1',
-    progress: progressMocks,
+  const {} = useAppletDetailsQuery(appletId, {
+    select: data => data.data.result,
   });
 
+  // const allAppletActivities = useMemo(() => {
```
anq83, line 28: Can you revert this file, please?
We'd integrate it when all api is done. No need to keep commented code.

anq83 (changes requested):

anq83 (approved):

PR #67, opened 2023-02-22
Added stacked slider component, moved Slider prop Types to separate f…

vmkhitaryanscn:
…ile, added initialValue oprtion

### DIFF
--- a/src/shared/ui/survey/StackedSlider.tsx
+++ b/src/shared/ui/survey/StackedSlider.tsx
@@ -1,0 +1,50 @@
+import { FC, useState } from 'react';
+
+import { StackedSliderProps } from '@shared/lib';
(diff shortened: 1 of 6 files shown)


src/shared/ui/survey/StackedSlider.tsx
```diff
@@ -0,0 +1,50 @@
...
+      onPress();
+    }
+  };
+
+  return (
+    <YStack>
+      {config.map((sliderConfig, index) => {
+        const { sliderLabel } = sliderConfig;
+        return (
+          <YStack key={`slider-${sliderLabel}`}>
```
BamMironov, line 31: It's better to set the `index` as a `key` in this case 'cause we could have the same label for many sliders

BamMironov (changes requested):

BamMironov (approved):

PR #79, opened 2023-03-03
M2-747(survey_items): added Stacked checkbox item

DimitrySakalouski:


src/shared/ui/survey/StackedCheckboxItem/StackedCheckboxItem.tsx
```diff
@@ -0,0 +1,54 @@
+import { FC } from 'react';
+import { StyleSheet } from 'react-native';
+
+import StackedCheckboxConfig from './types';
+import { CheckBox } from '../..';
```
vmkhitaryanscn, line 5: Something is wrong here

src/shared/ui/survey/StackedCheckboxItem/StackedCheckboxItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { StackedItemsGrid } from '../StackedItemsGrid';
+import { StackedRowItemValue } from '../StackedItemsGrid/types';
+
+type Props = {
+  value: Array<string[]>;
+  onChange: (value: Array<string[]>) => unknown;
```
vmkhitaryanscn, line 11: maybe void?

src/shared/ui/survey/StackedCheckboxItem/StackedCheckboxItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+  config: StackedCheckboxConfig;
+};
+
+const StackedCheckboxItem: FC<Props> = ({ value = [], onChange, config }) => {
```
vmkhitaryanscn, line 15: value = [],
this has to be required prop, I think there is no need for default value.

src/shared/ui/survey/StackedCheckboxItem/StackedCheckboxItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const onValueChange = (option: StackedRowItemValue, itemIndex: number) => {
+    if (!value[itemIndex]?.length) {
+      value[itemIndex] = [];
+    }
+
+    if (value[itemIndex].includes(option.name.en)) {
+      value[itemIndex] = value[itemIndex].filter(
+        valueItem => valueItem !== option.name.en,
```
vmkhitaryanscn, line 23: I think this has to be name 
In refactored backend there will be name : string instead of name: { en: string }.

vmkhitaryanscn (changes requested):

src/shared/ui/survey/StackedCheckboxItem/StackedCheckboxItem.tsx
DimitrySakalouski: Fixed

src/shared/ui/survey/StackedCheckboxItem/StackedCheckboxItem.tsx
DimitrySakalouski: But what if user takes survey for a first time? There would be no value, right?

src/shared/ui/survey/StackedCheckboxItem/StackedCheckboxItem.tsx
DimitrySakalouski: Fixed

src/shared/ui/survey/StackedCheckboxItem/StackedCheckboxItem.tsx
BamMironov: @vmkhitaryanscn  `value` is already required according to `Props`.

BamMironov (approved):

vmkhitaryanscn (approved):

PR #86, opened 2023-03-09
Added/configured react-native-audio-recorder-player and react-native-…

vmkhitaryanscn:
Added/configured react-native-audio-recorder-player and react-native-mime-types packages , added markdown audio support , fixed subscript and superscript fontSizes, created AudioPlayer component

Added MaterialIcons Fonts

Added react-native-- webview, autoheight-webview,video,video-player packages, added video, youtube video, html_block markdown support, added VideoPlayer component


src/shared/ui/survey/AudioPlayer.tsx
```diff
@@ -0,0 +1,80 @@
...
+    }
+  };
+
+  const play = async () => {
+    setPlaying(true);
+
+    await audioRecorderPlayer.startPlayer(uri);
+    audioRecorderPlayer.addPlayBackListener(data => {
+      if (data.currentPosition === data.duration) {
+        setPlaying(false);
```
BamMironov, line 43: We should probably call the `stop` function here also

BamMironov (changes requested):

src/shared/ui/survey/AudioPlayer.tsx
vmkhitaryanscn: @BamMironov I think when this if block is entered the player stops by itself, it doesn't loop.

src/shared/lib/markDownRules.tsx
```diff
@@ -141,15 +225,39 @@ const markDownRules: RenderRules = {
...
-          style={localStyles.image}
-          source={node.attributes.src}
+          mediaPlaybackRequiresUserAction
+          source={{
+            uri,
+          }}
         />
-      </>
+      );
+    }
```
BamMironov, line 252: Please move the logic to YoutubeVideo component so that you can use here like this:
```
else if (src?.includes('youtu')) {
  return <YoutubeVideo uri={src} key={node.key} />
}
```

src/shared/lib/markDownRules.tsx
```diff
@@ -84,6 +90,83 @@ export const activityMarkDownStyles = StyleSheet.create({
...
+
+  td {
+    font-size: 14px;
+    text-align: left;
+  }
+`;
+
+const onHtmlBlockLinkPress = (request: { url: string }) => {
+  if (request.url !== 'about:blank') {
+    Linking.openURL(request.url).then();
```
BamMironov, line 164: Please remove `.then()`

src/shared/lib/markDownRules.tsx
```diff
@@ -141,15 +225,39 @@ const markDownRules: RenderRules = {
...
-      <>
-        <CachedImage
+    const src = node.attributes?.src;
+    const mimeType = mime.lookup(src) || '';
+
+    if (mimeType.startsWith('audio/')) {
+      return <AudioPlayer uri={src} title={node.content} key={node.key} />;
+    } else if (mimeType.startsWith('video/') || src?.includes('.quicktime')) {
+      return <VideoPlayer uri={src} key={node.key} />;
+    } else if (src?.includes('youtu')) {
```
BamMironov, line 235: It is better to create variables for all the conditions:
```
const isAudio = mimeType.startsWith('audio/');
const isVideo = mimeType.startsWith('video/') || src?.includes('.quicktime');
const isYoutubeVideo = src?.includes('youtu');
```

src/shared/ui/survey/AudioPlayer.tsx
BamMironov: Ok

BamMironov (changes requested):

BamMironov (approved):

PR #90, opened 2023-03-15
M2-811: [Mobile] Migrate Flanker for Android

anq83:
M2-811: [Mobile] Migrate Flanker for Android


src/entities/flanker/lib/types/configuration.ts
```diff
@@ -0,0 +1,84 @@
...
+  buttonLabel: string;
+  minimumAccuracy: number;
+  continueText: Array<string>;
+  restartText: Array<string>;
+};
+
+export type StimulusScreen = {
+  id: StimulusConfigId;
+  correctChoice: number;
+  stimulus: { en: string };
```
vmkhitaryanscn, line 67: Does this need to be updated to stimulus: string from backend?

anq83: Unfortanetelly this contract for the existing html page which is working with PsichoJs library.

src/entities/flanker/ui/mocks/FlankerWithImageOnButton.ts
```diff
@@ -0,0 +1,118 @@
...
+  showFixation: true,
+  showFeedback: true,
+  showResults: true,
+  samplingMethod: 'fixed-order',
+  nextButton: 'OK',
+  sampleSize: 1,
+  trialDuration: 5000,
+  fixationDuration: 1000,
+  fixationScreen: {
+    value: 'Без названия.png',
```
BamMironov, line 111: Are you sure about that?

src/entities/flanker/ui/HtmlFlanker.tsx
```diff
@@ -0,0 +1,101 @@
+/* eslint-disable react-hooks/exhaustive-deps */
+/* eslint-disable react-native/no-inline-styles */
```
BamMironov, line 2: Why is it necessary to disable them?

BamMironov (changes requested):

src/entities/flanker/ui/mocks/FlankerWithImageOnButton.ts
anq83: Yes, just a mock, but based on real data taken from old app api

PR #91, opened 2023-03-15
Implemented replaceTextWithScreenVariables function, added MarkdownVa…

vmkhitaryanscn:
…riableReplacer class, added unit tests for these

### DIFF
--- a/src/shared/lib/utils/survey.ts
+++ b/src/shared/lib/utils/survey.ts
@@ -1,3 +1,5 @@
+import { format } from 'date-fns';
+
 import { colors } from '@shared/lib/constants';
 
 export const invertColor = (hex: string) => {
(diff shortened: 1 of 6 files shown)


src/shared/lib/utils/survey.ts
```diff
@@ -13,4 +15,208 @@ export const invertColor = (hex: string) => {
...
+
+type ActivityItems = {
+  variableName: string;
+  inputType: InputAnswer['inputType'];
+  valueConstraints: {
+    itemList: { name: string; value: string | number }[];
+  };
+}[];
+
+export class MarkdownVariableReplacer {
```
BamMironov, line 100: Please put it in a separate file.
You can create the "src/shared/survey" folder and move there everything that you created here.

src/shared/lib/utils/survey.ts
```diff
@@ -13,4 +15,208 @@ export const invertColor = (hex: string) => {
...
+  private markdown = '';
```
BamMironov, line 101: I think this should not be private property.

The `MarkdownVariableReplacer` must accept two arguments: `activityItems` and `answers`.
Then, the `process` function must accept one argument: `markdown`.
This way, we can create a replacer based on a context (activityItems, answers) and then process any markdown with it.

src/shared/lib/utils/survey.ts
```diff
@@ -13,4 +15,208 @@ export const invertColor = (hex: string) => {
...
+    );
+    const answerNotFound = foundIndex < 0 || !this.answers[foundIndex];
+
+    if (answerNotFound) {
+      return `[[${variableName}]]`;
+    }
+
+    const activityItem = this.activityItems[foundIndex];
+    let updated = '';
+    const inputAnswer: InputAnswer = <InputAnswer>{
```
BamMironov, line 183: Is it necessary to put <InputAnswer> generic here?

BamMironov (changes requested):

src/shared/lib/utils/survey.ts
vmkhitaryanscn: @BamMironov yes, it gives

TS2322: Type '{ inputType: "timeRange" | "date" | "text" | "radio" | "ageSelector" | "slider" | "checkbox"; answer: Answer; }' is not assignable to type 'InputAnswer'.   Type '{ inputType: "timeRange" | "date" | "text" | "radio" | "ageSelector" | "slider" | "checkbox"; answer: Answer; }' is not assignable to type 'CheckBoxInput'.     Types of property 'inputType' are incompatible.       Type '"timeRange" | "date" | "text" | "radio" | "ageSelector" | "slider" | "checkbox"' is not assignable to type '"checkbox"'.         Type '"timeRange"' is not assignable to type '"checkbox"

BamMironov (approved):

PR #96, opened 2023-03-20
M-860: added Photo Item

DimitrySakalouski:


ios/MindloggerMobile/Info.plist
```diff
@@ -37,8 +37,14 @@
...
 	</dict>
+	<key>NSCameraUsageDescription</key>
+	<string>MindLogger would like to use your camera</string>
+	<key>NSPhotoLibraryUsageDescription</key>
+  	<string>MindLogger would like access to your photo gallery</string>
 	<key>NSLocationWhenInUseUsageDescription</key>
 	<string>MindLogger would like to use your location when you are using the app</string>
+	<key>NSMicrophoneUsageDescription</key>
+	<string>MindLogger would like to use your microphone (for videos)</string>
 	<key>UIAppFonts</key>
```
vmkhitaryanscn, line 48: I think we better remove "(for videos)".

src/shared/lib/utils/galleryPermissions.ts
```diff
@@ -0,0 +1,47 @@
...
+  return await Permissions.request(GALLERY_IOS_PERMISSIONS);
+};
+
+export const checkGalleryPermissions = () =>
+  IS_ANDROID ? checkGalleryAndroidPermissions() : checkGalleryIOSPermissions();
+
+export const requestGalleryPermissions = async () =>
+  IS_ANDROID
+    ? await requestGalleryAndroidPermissions()
+    : await requestGalleryIOSPermissions();
```
BamMironov, line 47: No need to put await here here.

As a whole, I would do the following: 
```
const requestGalleryPermissions =
  IS_ANDROID
    ? requestGalleryAndroidPermissions
    : requestGalleryIOSPermissions;
```

BamMironov: The same you could do in `checkGalleryPermissions` function

src/shared/ui/survey/MediaItems/types.ts
```diff
@@ -0,0 +1,9 @@
+type MediaValue = {
+  uri: string;
+  filename: string;
```
vmkhitaryanscn, line 3: Does this naming come somewhere from package or some other place that doesn't depend on us? If no, we better rename to file**N**ame.

src/shared/lib/hooks/useCameraPermissions.ts
vmkhitaryanscn, line 1: @DimitrySakalouski @BamMironov should we place all permission related hooks, constants, and maybe constants to "shared/.../permissions" directory?

src/shared/ui/survey/MediaItems/PhotoItem.tsx
```diff
@@ -0,0 +1,60 @@
...
+  value?: MediaValue;
+};
+
+const PhotoItem: FC<Props> = ({ onChange, value }) => {
+  const { isCameraAccessGranted } = useCameraPermissions();
+  const { isGalleryAccessGranted } = useGalleryPermissions();
+
+  const handlePickImage = (response: any) => {
+    onChange(response);
+  };
```
BamMironov, line 30: You don't need this function.
Just use `onChange` on `launchImageLibrary` and `launchCamera` function calls

BamMironov (changes requested):

vmkhitaryanscn (approved):

src/shared/ui/survey/MediaItems/types.ts
BamMironov: @vmkhitaryanscn  filename is ok.
Source: https://dictionary.cambridge.org/dictionary/english/filename

src/shared/lib/hooks/useCameraPermissions.ts
BamMironov: Perfect idea. Let's do that.
@DimitrySakalouski  please move them to shared/lib/permission folder

ios/MindloggerMobile/Info.plist
DimitrySakalouski: I've copied that from legacy app so it seems to be correct

DimitrySakalouski: But in case we shall have pure mic usage then yes, you're right - I'll fix that

src/shared/ui/survey/MediaItems/PhotoItem.tsx
```diff
@@ -17,29 +22,49 @@ import MediaInput from './MediaInput';
 import MediaValue from './types';
 
 type Props = {
-  onChange: (value: MediaValue) => void;
+  onChange: (value: MediaValue) => void | Callback;
```
BamMironov, line 25: Why do need to return Callback here?

src/shared/ui/survey/MediaItems/PhotoItem.tsx
```diff
@@ -17,29 +22,49 @@ import MediaInput from './MediaInput';
...
   value?: MediaValue;
 };
 
 const PhotoItem: FC<Props> = ({ onChange, value }) => {
   const { isCameraAccessGranted } = useCameraPermissions();
   const { isGalleryAccessGranted } = useGalleryPermissions();
 
-  const handlePickImage = (response: any) => {
-    onChange(response);
+  const handlePickImage = (
```
BamMironov, line 33: Can we name it just `pickImage`?

BamMironov (changes requested):

PR #105, opened 2023-03-24
feat(about_screen): add about text for Applet About screen from api

DimitrySakalouski:


src/screens/ui/AboutAppletScreen.tsx
```diff
@@ -1,26 +1,45 @@
...
+      );
+    }
 
-  if (content.startsWith('404:')) {
-    content =
-      '# ¯\\\\_(ツ)_/¯ ' +
-      '\n # \n The authors of this applet have not provided any information!';
-  }
+    return appletAbout;
+  }, [appletAbout]);
```
BamMironov, line 36: There is no need to use useMemo here.

src/screens/ui/AboutAppletScreen.tsx
```diff
@@ -1,26 +1,45 @@
...
 import { MarkdownView } from '@shared/ui';
 import { YStack, ScrollView, Box } from '@shared/ui';
 
+type RouteProps = {
+  key: string;
+  name: string;
+  params: {
+    appletId: string;
+  };
+};
```
BamMironov, line 16: The types for screens should be declared in `screens/lib/types.ts`
Please take a look at `InProgressActivityScreen.tsx`. There you can find an example of how to connect type the right way.

BamMironov (changes requested):

BamMironov (approved):

PR #106, opened 2023-03-24
Feat/m2 988

BamMironov:


src/widgets/survey/ui/Intermediate.tsx
```diff
@@ -56,13 +56,23 @@ function Intermediate({
     select: r => r.data.result.activityFlows.find(o => o.id === flowId),
   });
 
+  let { data: allActivities } = useAppletDetailsQuery(appletId, {
+    select: r => r.data.result.activities,
```
anq83, line 60: Please do not use dto-s in the components directly. When you do so - you're merging data transfer layer with ui layer. Use the mapper from mu PR re. integration with mocks

anq83 (changes requested):

anq83 (approved):

PR #118, opened 2023-04-03
Feat/m2 1106

BamMironov:
Number selector pipeline item + additional text input + scroll button


src/features/pass-survey/model/hooks/useActivityStepper.ts
```diff
@@ -11,8 +11,14 @@ function useActivityStepper(state: ActivityState | undefined) {
...
   const isLastStep = items && step === items.length - 1;
 
+  const additionalAnswerRequired =
+    currentPipelineItem?.additionalText?.required;
+
   const canMoveNext =
-    isTutorialStep || currentPipelineItem?.isSkippable || !!answers[step];
+    isTutorialStep ||
+    currentPipelineItem?.isSkippable ||
+    (!!answers[step]?.answer &&
```
anq83, line 20: Move to variables:
const hasAnswer =  !!answers[step]?.answer
const hasAdditionalAnswer =  !!answers[step]?.additionalAnswer

src/shared/ui/survey/NumberSelector.tsx
```diff
@@ -2,27 +2,27 @@ import { FC, useMemo } from 'react';
...
+  config: NumberSelectorConfig;
+  value: string;
+  onChange: (value: string) => void;
 };
 
-const AgeSelector: FC<Props> = ({ config, value, onChange }) => {
+const NumberSelector: FC<Props> = ({ config, value, onChange }) => {
   const { t } = useTranslation();
 
   const ageRange = useMemo(
```
anq83, line 21: ageRange rename to numberRange

anq83 (changes requested):

src/features/pass-survey/ui/ActivityItem.tsx
```diff
@@ -1,35 +1,59 @@
...
 import { SurveySlider } from '@shared/ui';
 
-import { PipelineItem, PipelineItemResponse } from '../lib';
+import AdditionalText from './AdditionalText';
+import { Answer, PipelineItem, PipelineItemResponse } from '../lib';
 import { TextResponseMapper } from '../model/responseMappers';
 
 type Props = {
-  value: any;
+  value: Answer | undefined;
```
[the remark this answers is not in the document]
BamMironov: @anq83 
In that case, it is not a mistake. Value is provided by the `answers` object and by default is empty. Hence, getting a `value` by the key from the object can result in an undefined value. Sure, I can pass Props like `value={answers[step] ?? null}` but this is nonsense.

Anyway, in my next PR, there will be no `Answer | undefined`. I am going with `value?: Answer;` type which suits perfectly in that case. So we can just ignore it for now.

src/shared/ui/survey/NumberSelector.tsx
BamMironov: right

src/features/pass-survey/ui/ActivityItem.tsx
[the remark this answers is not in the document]
BamMironov: Okay, I will do it right now

anq83 (approved):

PR #152, opened 2023-04-19
M2-1468 M2-1450 M2-1448

anq83:
M2-1468 [Mobile] Update activity group list every minute
M2-1450 [Mobile] Logout - react query cache clean up
M2-1448 [Mobile] Consider isHidden activities and flows in activity groups list

### DIFF
--- a/src/widgets/activity-group/model/hooks/useTimer.ts
+++ b/src/widgets/activity-group/model/hooks/useTimer.ts
@@ -1,0 +1,24 @@
+import { useCallback, useEffect, useState } from 'react';
+
+import { AppClockTimer } from '@app/shared/lib';
+
+const useTimer = () => {
+  const [, reRender] = useState(false);
+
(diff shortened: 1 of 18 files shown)


src/widgets/activity-group/model/hooks/useTimer.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  useEffect(() => {
+    const timer = new AppClockTimer(onTick, false);
+
+    timer.start();
+
+    return () => {
+      timer.stop();
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
BamMironov, line 20: You can remove the comment and move the `onTick` function inside the `useEffect` scope. No `eslint` warnings are expected in that case

BamMironov (changes requested):

BamMironov (approved):

PR #154, opened 2023-04-19
Fixed M2-1156,M2-1440,M2-1442,M2-1444, M2-1438, M2-1146

vmkhitaryanscn:
https://mindlogger.atlassian.net/browse/M2-1156
https://mindlogger.atlassian.net/browse/M2-1440
https://mindlogger.atlassian.net/browse/M2-1442
https://mindlogger.atlassian.net/browse/M2-1444
https://mindlogger.atlassian.net/browse/M2-1438
https://mindlogger.atlassian.net/browse/M2-1146

### DIFF
--- a/src/features/pass-survey/model/pipelineBuilder.ts
+++ b/src/features/pass-survey/model/pipelineBuilder.ts
@@ -236,12 +236,14 @@ export function buildPipeline(activity: ActivityDetails): PipelineItem[] {
       return Array.isArray(item) ? [...items, ...item] : [...items, item];
     }, [])
     .map(item => {
+      const isAbleToMoveBack =
+        activity.responseIsEditable && item.isAbleToMoveBack;
+      const isSkippable = activity.isSkippable || item.isSkippable;
+
       return {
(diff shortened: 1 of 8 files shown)


src/features/pass-survey/model/pipelineBuilder.ts
```diff
@@ -234,7 +234,17 @@ export function buildPipeline(activity: ActivityDetails): PipelineItem[] {
     })
     .reduce<PipelineItem[]>((items, item) => {
       return Array.isArray(item) ? [...items, ...item] : [...items, item];
-    }, []);
+    }, [])
+    .map((item, index) => {
+      return {
+        ...item,
+        isAbleToMoveBack:
+          !activity.responseIsEditable && index !== 0
```
BamMironov, line 242: Please remove or modify the `index !== 0` condition because we also need to take into account that the item is not "Splash".
Also, I would recommend moving the complex condition into a separate variable.

BamMironov (changes requested):

PR #172, opened 2023-04-27
Fixes M2-1591, M2-1588

BamMironov:
https://mindlogger.atlassian.net/browse/M2-1591
https://mindlogger.atlassian.net/browse/M2-1588


src/widgets/survey/model/hooks/useFlowRecordInitialization.ts
```diff
@@ -0,0 +1,79 @@
...
+      const activityIds = applet.activityFlows.find(
+        flow => flow.id === flowId,
+      )?.activityIds;
+
+      return buildActivityFlowPipeline({
+        activityIds: activityIds!,
+        appletId,
+        eventId,
+        flowId,
+        startFrom: step,
```
anq83, line 58: Will step always equial to 0? If so, then set it explicitly and remove dependency

src/widgets/survey/ui/Finish.tsx
```diff
@@ -41,21 +41,30 @@ function FinishItem({
       appletId,
       activityId,
       eventId,
+      order,
     });
 
+  const { isTimerElapsed } = useFlowState({
```
anq83, line 47: This flag will not work correctly due to double usage of useState inside it

src/widgets/survey/ui/Finish.tsx
```diff
@@ -41,21 +41,30 @@ function FinishItem({
...
+
   const {
     mutate: sendAnswers,
     isLoading: isSendingAnswers,
     isSuccess: successfullySentAnswers,
     isPaused: isOffline,
   } = useActivityAnswersMutation({
     onError: error => {
-      if (error.evaluatedMessage) {
+      if (error.response.status !== 401 && error.evaluatedMessage) {
```
anq83, line 60: Will it be redirected to login screen exactly at this moment ?

src/widgets/survey/ui/Intermediate.tsx
```diff
@@ -60,13 +63,23 @@ function Intermediate({
     select: r => mapActivitiesFromDto(r.data.result.activities),
   });
 
-  const totalActivities = activityFlow!.activityIds.length;
```
anq83, line 63: useAppletDetailsQuery above should be excluded and all dependent code should relay only on pipleline
add a comment - todo at least

anq83 (changes requested):

src/widgets/survey/model/hooks/useFlowRecordInitialization.ts
BamMironov: No, it won't always equal 0

src/widgets/survey/ui/Finish.tsx
BamMironov: Automatic redirection to the Logic screen is handled on `RootNavigator.tsx`
image

src/widgets/survey/ui/Intermediate.tsx
BamMironov: Added a comment

anq83 (approved):

PR #175, opened 2023-04-27
Feat/m2 1541 stacked checkbox

vmkhitaryanscn:
added - https://jira.scnsoft.com/projects/CMIML/issues/CMIML-1017
+
fixed bug - https://mindlogger.atlassian.net/jira/software/c/projects/M2/issues/M2-1700

### DIFF
--- a/src/shared/ui/survey/StackedCheckboxItem/StackedCheckboxItem.tsx
+++ b/src/shared/ui/survey/StackedCheckboxItem/StackedCheckboxItem.tsx
@@ -54,43 +54,39 @@ const StackedCheckboxItem: FC<Props> = ({
   }, [rows, textReplacer]);
 
   const onValueChange = (option: StackedRowItemValue, row: Item) => {
-    const rowId = row.id;
+    const currentRowId = row.id;
     const optionId = option.id;
-    const rowInitialValue = {
-      rowId,
-      optionIds: [optionId],
-    };
 
     if (!values) {
-      const value = [rowInitialValue];
+      const newValues = [
+        {
+          rowId: currentRowId,
+          optionIds: [optionId],
+        },
+      ];
 
-      onChange(value);
+      onChange(newValues);
     } else {
-      const clonedValues = [...values];
-      let foundRow = false;
+      const currentRow = values.find(({ rowId }) => rowId === currentRowId);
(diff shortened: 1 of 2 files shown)


src/shared/ui/Tooltip.tsx
```diff
@@ -10,11 +10,14 @@ import { MarkdownView, ScrollView, YStack } from '../ui';
 
 type TooltipProps = {
   children: React.ReactNode;
-  markdown?: string;
+  markdown?: string | null;
```
DimitrySakalouski, line 13: In case if B.E. will send null as empty tooltip then there is no need to have "?" ("undefined")

src/shared/ui/survey/StackedItemsGrid/StackedItemsGrid.tsx
```diff
@@ -1,76 +1,66 @@
...
-                style={styles.image}
-                resizeMode="contain"
-                source={option.image}
-              />
-            )}
-          </Center>
-        </Tooltip>
+        <Center>
+          {option.tooltip ? (
+            <Tooltip markdown={option.tooltip}>
```
BamMironov, line 43: Variable replacer is missing

BamMironov (changes requested):

src/shared/ui/survey/StackedCheckboxItem/StackedCheckboxItem.tsx
```diff
@@ -1,46 +1,133 @@
...
+              id => id !== optionId,
+            );
+          } else {
+            currentRow.optionIds.push(optionId);
+          }
+        }
+      }
+      if (!foundRow) {
+        clonedValues.push(rowInitialValue);
+      }
```
BamMironov, line 87: Instead of this for-loop and if-statement, you can do something like these:
```
const row = value[rowId] ?? {};
const optionIds = row?.optionIds ?? [];

const newOptionIds = [...optionIds];

if (optionIds.includes(optionId)) {
  newOptionIds = optionIds.filter(
    id => id !== optionId,
  );
} else {
  newOptionIds = [...newOptionIds, optionId];
}

const newValue = {
  ...values,
  [rowId]: newOptionIds,
};

```

BamMironov (changes requested):

PR #204, opened 2023-05-16
fix(survey_items): video item alert issue

ghost:


src/shared/ui/survey/MediaItems/MediaInput.tsx
```diff
@@ -44,9 +55,13 @@ const MediaInput: FC<Props> = ({
   };
 
   return (
-    <TouchableOpacity onPress={onUploadPress}>
-      <UploadButton>{children}</UploadButton>
-    </TouchableOpacity>
+    <UploadButton>
```
anq83, line 58: It seems UploadButton is just an area where internals placed. This not a button. Please rename to something like "ContentWrapper"

src/shared/ui/survey/MediaItems/MediaInput.tsx
```diff
@@ -27,6 +37,7 @@ const MediaInput: FC<Props> = ({
   mode,
   onOpenCamera,
   onShowMediaLibrary,
+  iconComponent,
```
anq83, line 40: rename to uploadIcon

anq83 (changes requested):

PR #210, opened 2023-05-19
Feature/answers encryption

vmkhitaryanscn:


src/shared/lib/encryption/encryption.ts
```diff
@@ -0,0 +1,110 @@
...
+  privateKey: number[];
+  appletPrime: number[];
+  appletBase: number[];
+};
+
+type GetAESKeyProps = {
+  userPrivateKey: number[];
+  appletPublicKey: number[];
+  appletPrime: number[];
+  appletBase: number[];
```
BamMironov, line 18: Let's not bind to the "applet" and "user" name. Please keep them without it, e.g. userPrivateKey => privateKey, appletPublicKey => publicKey, etc.

That concerns type GetPublicKeyProps as well

src/shared/lib/encryption/encryption.ts
```diff
@@ -0,0 +1,110 @@
...
+  public getPrivateKey = ({
+    userId,
+    email,
+    password,
+  }: GetPrivateKeyProps): number[] => {
+    const key1 = crypto
+      .createHash('sha512')
+      .update(password + email)
+      .digest();
+    const key2 = crypto
```
BamMironov, line 34: Please rename:
key1 => leftPart
key2 => rightPart

src/shared/lib/encryption/encryption.ts
```diff
@@ -0,0 +1,110 @@
...
+  userPrivateKey: number[];
+  appletPublicKey: number[];
+  appletPrime: number[];
+  appletBase: number[];
+};
+
+type EncryptDataProps = { text: string; key: number[] };
+type DecryptDataProps = { text: string; key: string };
+
+class Encryption {
```
BamMironov, line 24: Please come up with a better name. Encryption means a process and cannot be interpreted as a class. EncryptionService or EncryptionManager, for example

src/shared/lib/hooks/useEncryption.ts
```diff
@@ -0,0 +1,70 @@
...
+  privateKey: number[];
+  appletPrime: number[];
+  appletBase: number[];
+};
+
+type GenerateAesKeyProps = {
+  userPrivateKey: number[];
+  appletPublicKey: number[];
+  appletPrime: number[];
+  appletBase: number[];
```
BamMironov, line 21: Same here

src/shared/lib/hooks/useEncryption.ts
```diff
@@ -0,0 +1,70 @@
...
+  userId: string;
+  email: string;
+  password: string;
+};
+
+type GenerateUserPublicKeyParams = {
+  privateKey: number[];
+  appletPrime: number[];
+  appletBase: number[];
+};
```
BamMironov, line 15: Same here

BamMironov (changes requested):

BamMironov (approved):

PR #218, opened 2023-05-26
Feature/answers encryption

vmkhitaryanscn:


src/entities/activity/api/useActivityAnswersMutation.ts
```diff
@@ -17,6 +17,9 @@ type SendAnswersInput = {
   createdAt: number;
   answers: AnswerDto[];
   appletEncryption: AppletEncryptionDTO;
+  itemIds: any[];
```
BamMironov, line 20: Should be string[]

src/widgets/survey/model/mappers.ts
```diff
@@ -1,46 +1,69 @@
...
-  return type === 'function' || (type === 'object' && !!obj);
-}
+type Answer = PipelineItemAnswer['value'];
 
 export function mapAnswersToDto(
   pipeline: PipelineItem[],
   answers: Answers,
-  flowId: string | null,
 ): Array<AnswerDto> {
   const filteredAnswers = Object.entries(answers).filter(([_, answer]) => {
```
BamMironov, line 14: Just a comment.
When conditional logic is merged we should all add another filter that filters items that cannot pass the conditional logic.

src/widgets/survey/model/mappers.ts
```diff
@@ -1,46 +1,69 @@
...
-    }
+      case 'Checkbox':
+        return convertToCheckboxAnswer(answer);
+
+      case 'Slider':
+        return convertToSliderAnswer(answer);
 
-    return dto;
+      case 'NumberSelect':
+        return null;
```
BamMironov, line 34: Why null?

src/widgets/survey/ui/Finish.tsx
```diff
@@ -103,22 +103,28 @@ function FinishItem({
...
-        // the timeout is set up, because the encryption process blocks loading the cached background image
-        sendAnswers({
-          appletId,
-          createdAt: getUnixTimestamp(Date.now()),
-          version: activityStorageRecord.appletVersion,
-          answers: answers,
-          appletEncryption,
-        });
-      }, 100);
+      const itemIds = Object.entries(activityStorageRecord.answers).map(
```
BamMironov, line 112: Just a comment.
When conditional logic is merged we should all add another filter that filters items that cannot pass the conditional logic.

BamMironov (changes requested):

BamMironov (approved):

PR #222, opened 2023-05-31
fix(survey_items): border color for media input item when media is selected

DimitrySakalouski:
fix(survey_items): border color for media input item when media is selected

https://mindlogger.atlassian.net/browse/M2-1904

### DIFF
--- a/src/shared/ui/survey/MediaItems/MediaInput.tsx
+++ b/src/shared/ui/survey/MediaItems/MediaInput.tsx
@@ -9,7 +9,7 @@ import {
 import { styled } from '@tamagui/core';
 import { useTranslation } from 'react-i18next';
 
-import { Center, Box } from '@shared/ui';
+import { Center, Box, BoxProps } from '@shared/ui';
 
 const styles = StyleSheet.create({
   touchable: {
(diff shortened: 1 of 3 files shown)


src/shared/ui/survey/MediaItems/MediaInput.tsx
```diff
@@ -61,7 +69,7 @@ const MediaInput: FC<Props> = ({
   };
 
   return (
-    <ContentWrapper height={windowWidth * 0.85}>
+    <ContentWrapper hasValue={hasValue} height={windowWidth * 0.85}>
```
BamMironov, line 72: Can we just use `borderColor` property right here? I don't think there is a need to create a variant for the sake of one style property.

BamMironov: Please make `MediaInput` accept BoxProps and just pass these styled props to `ContentWrapper`.
You will decide which color is passed to `MediaInput` in the parent component (PhotoItem, VideoItem)

BamMironov (changes requested):

BamMironov (approved):

PR #223, opened 2023-05-31
Added system variables text replacing functionality

vmkhitaryanscn:


src/features/pass-survey/lib/markdownVariableReplacer.ts
```diff
@@ -64,11 +71,78 @@ export class MarkdownVariableReplacer {
...
+        /\[Time_Elapsed_Activity_Last_Completed\]/gi,
+        this.formatTimeElapsed(),
+      )
+      .replaceAll(
+        /\[Time_Activity_Last_Completed\]/gi,
+        this.formatLastResponseTime(),
+      );
+  };
+
+  private formatTimeElapsed = () => {
```
BamMironov, line 112: Although the function does some formatting stuff inside, the main purpose of it is to return a string.
So please consider another naming, like getTimeElapsed.
Format functions imply that they accept a string as an input (parameter) and return a string in a desired format as an output.

BamMironov (changes requested):

anq83 (approved):

PR #228, opened 2023-06-02
codespell: config, workflow and some typo fixes

yarikoptic:


src/shared/ui/Splash.tsx
```diff
@@ -52,7 +52,7 @@ const SplashScreen = ({ isAppReady }: { isAppReady: boolean }) => {
 
   return (
     <Animated.View
-      collapsable={false}
+      collapsible={false}
```
yarikoptic, line 55: not sure if this is smth to be fixed in some library interface upstream?

BamMironov:
Sorry, this is a bit outdated at this point. Moreover, some changes are quite risky as renaming project internals is always a risky action and should be well-tested by developers, and unfortunately, we do not have reserved time for it

yarikoptic:
IMHO your loss, but let me know if you would like me to update -- easy to do

PR #256, opened 2023-06-27
FEAT(survey_items): drawing item adjustments

DimitrySakalouski:


src/entities/drawer/ui/DrawingTest.tsx
```diff
@@ -43,7 +43,7 @@ const DrawingTest: FC<Props> = props => {
   const handleToggle = () => {
     !isDrawingActive && scrollToEnd();
 
-    toggleScrollRef.current(false);
+    toggleScrollRef.current(isDrawingActive);
```
anq83, line 46: This PR doesn't show correctly all affected files (6), maybe because we reverted it yesterday, so, please create a new PR from the 6 commits in the backup branch (origin/m2-2188-drawer-keep-before-revert) by cherry-picking them.

anq83 (changes requested):

PR #258, opened 2023-06-27
M2 2188 drawing item

DimitrySakalouski:
Also added fixes for https://mindlogger.atlassian.net/browse/M2-2410

### DIFF
--- a/src/shared/ui/ScrollableContent.tsx
+++ b/src/shared/ui/ScrollableContent.tsx
@@ -10,6 +10,7 @@ import {
   NativeScrollEvent,
   NativeSyntheticEvent,
   StyleSheet,
+  useWindowDimensions,
 } from 'react-native';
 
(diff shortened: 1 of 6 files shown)


src/shared/ui/ScrollableContent.tsx
```diff
@@ -54,6 +64,15 @@ const ScrollableContent: FC<Props> = ({ children, scrollEnabled }: Props) => {
     setEndOfContentReachedOnce(true);
   }
 
+  const isAreaScrollable = useMemo(() => {
+    return !!(scrollContentHeight && scrollContentHeight >= windowHeight);
+  }, [scrollContentHeight, windowHeight]);
```
BamMironov, line 69: There is no need to use useMemo here

BamMironov (changes requested):

anq83 (approved):

BamMironov (approved):

PR #270, opened 2023-07-06
Feat/firebase

BamMironov:
Integration with Firebase + FCM token management


src/shared/lib/records/SystemRecord.ts
```diff
@@ -0,0 +1,22 @@
+import { createStorage } from '@shared/lib/storages';
+
+const storage = createStorage('system');
+
+const DEVICE_ID_KEY = 'deviceId';
```
DimitrySakalouski, line 5: Maybe it is not so important but can we move it to "constants" file

DimitrySakalouski (approved):

PR #271, opened 2023-07-06
Fixed applets refreshing bug, when no internet connection available

vmkhitaryanscn:


src/features/applets-refresh/ui/AppletsRefresh.tsx
```diff
@@ -15,6 +15,7 @@ type Props = Omit<RefreshControlProps, 'refreshing' | 'onRefresh'>;
 
 const AppletsRefresh: FC<Props> = props => {
   const queryClient = useQueryClient();
+  const [appletsRefreshing, setAppletsRefreshing] = useState(false);
```
anq83, line 18: I'll have to check this fix

anq83 (changes requested):

BamMironov (approved):

src/features/applets-refresh/ui/AppletsRefresh.tsx
anq83: I've checked on ios simulator. The spenner is not shown now, when pull-to-refresh.

anq83 (changes requested):

PR #284, opened 2023-07-19
M2-2633: [Mobile app] Data Export: Drawing canvas is not 100x100

anq83:
https://mindlogger.atlassian.net/browse/M2-2633

### DIFF
--- a/src/entities/drawer/ui/DrawingBoard.tsx
+++ b/src/entities/drawer/ui/DrawingBoard.tsx
@@ -27,8 +27,6 @@ import {
   DrawLine,
   DrawPoint,
   Point,
-  transformByWidth,
-  transformBack,
   ResponseSerializer,
   DrawResult,
   CachedBezierItem,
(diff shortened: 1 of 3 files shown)


src/entities/drawer/ui/DrawingBoard.tsx
```diff
@@ -64,23 +62,23 @@ const DrawingBoard: FC<Props> = props => {
 
   const pathsCache = useRef<Array<SkPath>>([]).current;
 
-  const normalizedLines = useMemo(() => {
-    return transformByWidth(value, width);
-  }, [value, width]);
+  const currentLines = useMemo(() => {
+    return [...value];
+  }, [value]);
```
BamMironov, line 67: I suppose we don't need to use useMemo here already in this case

BamMironov (changes requested):

src/entities/drawer/ui/DrawingBoard.tsx
anq83: Then we'd get a new instance of array every render

BamMironov (approved):

PR #288, opened 2023-07-21
fix(app_infrastructure): remove lanscape mode for ipad

DimitrySakalouski:


BamMironov:
@DimitrySakalouski  please also add this
https://stackoverflow.com/a/32728607

BamMironov (approved):

PR #293, opened 2023-07-24
Added client meta data to send answer request, handling media files i…

vmkhitaryanscn:
…n userActions, also removing unused media items from it

M2-2650
M2-2638

### DIFF
--- a/src/entities/activity/lib/services/AnswersUploadService.ts
+++ b/src/entities/activity/lib/services/AnswersUploadService.ts
@@ -6,6 +6,7 @@ import {
   AnswerService,
   FileService,
   ObjectAnswerDto,
+  UserActionDto,
 } from '@app/shared/api';
 import { MediaFile } from '@app/shared/ui';
 import { UserPrivateKeyRecord } from '@entities/identity/lib';
@@ -296,6 +297,7 @@ class AnswersUploadService implements IAnswersUploadService {
         identifier,
       },
       createdAt: data.createdAt,
+      client: data.client,
     };
 
     return encryptedData;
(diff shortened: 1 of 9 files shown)


src/entities/activity/lib/services/AnswersUploadService.ts
```diff
@@ -304,14 +306,72 @@ class AnswersUploadService implements IAnswersUploadService {
...
+        !userActionValue?.uri ||
+        originalAnswerIndex === -1
+      ) {
+        return userAction;
+      }
+
+      return {
+        ...userAction,
+        response: {
+          value: (updatedAnswers[originalAnswerIndex] as ObjectAnswerDto).value,
```
anq83, line 352: I suppose in case of svg - here will be the full object which also contains lines array, it is correct?

src/entities/activity/lib/services/AnswersUploadService.ts
```diff
@@ -304,14 +306,72 @@ class AnswersUploadService implements IAnswersUploadService {
...
+      const response = userAction?.response as ObjectAnswerDto;
+      const userActionValue = response?.value as MediaFile;
+      const originalAnswerIndex = originalAnswers.findIndex(answer => {
+        const currentAnswerValue = (answer as ObjectAnswerDto)
+          ?.value as MediaFile;
+        return currentAnswerValue?.uri === userActionValue?.uri;
+      });
+
+      if (
+        userAction.type !== 'SET_ANSWER' ||
```
anq83, line 342: can you split this if to 2 if-s, leave here the one which checks only index, the 1st move at the top of this function

src/entities/activity/lib/services/AnswersUploadService.ts
```diff
@@ -304,14 +306,72 @@ class AnswersUploadService implements IAnswersUploadService {
...
+    const userActions = modifiedBody.userActions;
+    const updatedAnswers = modifiedBody.answers;
+
+    const updatedUserActions = userActions.map((userAction: UserActionDto) => {
+      const response = userAction?.response as ObjectAnswerDto;
+      const userActionValue = response?.value as MediaFile;
+      const originalAnswerIndex = originalAnswers.findIndex(answer => {
+        const currentAnswerValue = (answer as ObjectAnswerDto)
+          ?.value as MediaFile;
+        return currentAnswerValue?.uri === userActionValue?.uri;
```
anq83, line 338: Is it possible when both arguments euqal to undefined? If so, then the index will be probably incorrect?

src/entities/activity/lib/services/AnswersUploadService.ts
```diff
@@ -304,14 +306,72 @@ class AnswersUploadService implements IAnswersUploadService {
...
+        response: {
+          value: (updatedAnswers[originalAnswerIndex] as ObjectAnswerDto).value,
+        },
+      };
+    });
+
+    const userActionsWithoutEmptyFiles = updatedUserActions.map(
+      (userAction: UserActionDto) => {
+        const response = userAction?.response as ObjectAnswerDto;
+        const userActionValue = response?.value as MediaFile;
```
anq83, line 360: Not sure it's correct as response.value is probably string (in case of image, video, audio), but it's also object in case of svg

anq83 (changes requested):

PR #298, opened 2023-07-26
Added userActions modifier after files successfully uploaded M2--2638

vmkhitaryanscn:


anq83 (changes requested):

src/entities/activity/lib/services/AnswersUploadService.ts
```diff
@@ -320,6 +321,13 @@ class AnswersUploadService implements IAnswersUploadService {
 
     const modifiedBody = await this.uploadAllMediaFiles(body);
 
+    const updatedUserActions = this.mapUserActionsMediaFilesToDto(
```
anq83, line 324: Plase think how to try-catch this function, maybe add an internal one and wrap it with try-catch.

anq83 (changes requested):

src/entities/activity/lib/services/AnswersUploadService.ts
anq83: Show me via teams, please when you will have a solution

anq83 (approved):

PR #299, opened 2023-07-26
M2-2650 Added meta client information in send answers request

vmkhitaryanscn:


src/shared/lib/constants/index.ts
```diff
@@ -7,6 +7,9 @@ import { PERMISSIONS } from 'react-native-permissions';
 
 import { Language } from '../types';
 
+const { width: VIEWPORT_WIDTH, height: VIEWPORT_HEIGHT } =
```
anq83, line 10: Will it always correct?  it's evaluated when app just started before any view creation

src/widgets/survey/lib/metaHelpers.ts
```diff
@@ -0,0 +1,15 @@
+import {
+  APP_VERSION,
+  META_APP_NAME,
+  VIEWPORT_HEIGHT,
+  VIEWPORT_WIDTH,
+} from '@shared/lib';
+
+const getMetaClientInformation = () => ({
```
anq83, line 8: Please rename to getClientInformation

anq83 (changes requested):

src/shared/lib/constants/index.ts
BamMironov: In React Native width and height dimensions can change in two scenarios:
1. The device is rotating and changes the orientation (not our case, we don't support horizontal orientation).
2. The device is foldable. Some android devices can fold but we don't support them for now anyway

BamMironov (approved):

BamMironov (approved):

PR #315, opened 2023-08-01
fix(survey_items): added container for table markdown view

DimitrySakalouski:
https://mindlogger.atlassian.net/browse/M2-2828


src/shared/lib/markDownRules.tsx
```diff
@@ -233,6 +233,15 @@ const markDownRules: RenderRules = {
       </Box>
     );
   },
+  table: (node, children) => {
+    const { width: screenWidth } = Dimensions.get('screen');
+
+    return (
+      <Box key={node.key} width={screenWidth - 32}>
```
BamMironov, line 240: It is not clear what is 32. Please create a variable that describes it. 
For example, const PADDING_X = 32

BamMironov (changes requested):

BamMironov (approved):

PR #318, opened 2023-08-02
fixes for M2-2791, M2-2792 - user actions for ab trails

anq83:
https://mindlogger.atlassian.net/browse/M2-2791

https://mindlogger.atlassian.net/browse/M2-2791

### DIFF
--- a/src/features/pass-survey/model/hooks/useUserActionManager.ts
+++ b/src/features/pass-survey/model/hooks/useUserActionManager.ts
@@ -25,7 +25,8 @@ function useUserActionManager({ activityId, activityState }: Args) {
 
   const updateUserAction = (action: UserAction) => {
     return activityState!.actions.map(o => {
-      return o.payload.activityItemId === action.payload.activityItemId
+      return o.payload.activityItemId === action.payload.activityItemId &&
+        o.type === action.type
         ? action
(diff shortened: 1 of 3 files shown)


src/features/pass-survey/model/hooks/useUserActionManager.ts
```diff
@@ -43,6 +44,7 @@ function useUserActionManager({ activityId, activityState }: Args) {
     const shouldUpdateLastAction =
       lastUserAction &&
       lastUserAction?.payload?.activityItemId === currentPipelineItem.id &&
+      lastUserAction.type === 'SET_ANSWER' &&
```
BamMironov, line 47: Why would we need this check?
updateUserActionsWithAnswer implies to be called only SET_ANSWER action

BamMironov (changes requested):

src/features/pass-survey/model/hooks/useUserActionManager.ts
anq83: Because AB Trails has 4 ab-trail activity items and 4 tutorials. Every totorial has the same id as the next ab-trail item. So, we need to update only ab-trail and do not affect tutorial.

PR #353, opened 2023-08-28
Fixed M2-3058 , added fullscreen functionality for videos on ios

vmkhitaryanscn:
Fixed M2-3058 , added fullscreen functionality for videos on ios, now thumbnails of videos will be shown on ios

### DIFF
--- a/src/shared/ui/survey/VideoPlayer.tsx
+++ b/src/shared/ui/survey/VideoPlayer.tsx
@@ -49,6 +49,12 @@ const VideoPlayer: FC<Props> = ({
     playerRef.current?.stop();
   };
 
+  useEffect(() => {
+    if (!autoPlay) {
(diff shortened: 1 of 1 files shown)


src/shared/ui/survey/VideoPlayer.tsx
```diff
@@ -49,17 +49,22 @@ const VideoPlayer: FC<Props> = ({
...
+  }, [autoPlay, playerRef]);
+
   return (
     <Center flex={1} w="100%">
       <VideoPlayerBase
         ref={playerRef}
         video={{
           uri,
         }}
-        autoplay={autoPlay}
```
BamMironov, line 59: Why does it not work?

BamMironov (changes requested):

BamMironov (approved):

PR #356, opened 2023-08-30
Fixed M2-3180, M2-3090 timer and idle timer bug fixes

vmkhitaryanscn:
[M2-3180](https://mindlogger.atlassian.net/browse/M2-3180)
[M2-3090](https://mindlogger.atlassian.net/browse/M2-3090)

### DIFF
--- a/src/shared/lib/timers/TimerBase.ts
+++ b/src/shared/lib/timers/TimerBase.ts
@@ -6,6 +6,12 @@ abstract class TimerBase {
   hasStarted: boolean = false;
 
   protected start(): void {
+    this.hasStarted = true;
+
+    if (this.listener) {
+      return;
+    }
+
     this.listener = AppState.addEventListener('change', nextAppState => {
       if (nextAppState === 'active') {
         if (this.onForeground) {
@@ -17,7 +23,6 @@ abstract class TimerBase {
         }
       }
     });
-    this.hasStarted = true;
   }
 
   abstract setTimer(duration?: number): void;


src/shared/lib/timers/AppTimer.ts
```diff
@@ -25,13 +25,13 @@ class AppTimer extends TimerBase {
     this.setTimer();
   }
 
-  stop() {
-    super.stop();
+  stop(isRestart = false) {
```
BamMironov, line 28: Would suggest a different approach. Please see the comment for TimerBase

src/shared/lib/timers/TimerBase.ts
```diff
@@ -25,12 +23,14 @@ abstract class TimerBase {
...
   protected onBackground?(): void;
 
-  public stop(): void {
+  public stop(isRestart = false): void {
     this.hasStarted = false;
-    this.removeAppStateListener();
+    if (!isRestart) {
+      this.removeAppStateListener();
+    }
   }
```
BamMironov, line 34: Instead of using `isRestrart` flag, we can add a new listener in `start` method instead. 
I mean, we can move the creation of a listener from `constructor` to `start`.

src/widgets/activity-group/model/factories/ActivityGroupsBuilder.ts
```diff
@@ -296,6 +296,13 @@ class ActivityGroupsBuilder implements IActivityGroupsBuilder {
         item.availableTo = MIDNIGHT_DATE;
       }
 
+      const eventTimer = event.timers?.timer;
```
BamMironov, line 299: Please rename it to `timeLeftToComplete` for more clarity

src/shared/lib/timers/TimerBase.ts
```diff
@@ -8,9 +8,7 @@ abstract class TimerBase {
   constructor() {
     this.listener = AppState.addEventListener('change', nextAppState => {
       if (nextAppState === 'active') {
-        if (this.onForeground) {
-          this.onForeground();
-        }
+        this.onForeground();
```
BamMironov, line 11: I think we can left it as is

BamMironov (changes requested):

PR #357, opened 2023-08-31
M2 2578 integrate data loss

anq83:


src/entities/activity/lib/services/AnswersUploadService.ts
```diff
@@ -144,12 +167,22 @@ class AnswersUploadService implements IAnswersUploadService {
...
       );
     }
   }
 
+  private logFilesUploadCheck(
+    uploadChecks: CheckFilesUploadResults,
+    position: '1' | '2',
+  ) {
+    this.logger.log(
+      `[UploadAnswersService.uploadAllMediaFiles] Check if files uploaded #${position}:\n\n` +
```
BamMironov, line 181: Please rename it to "...Check if **log files** uploaded..."

src/shared/lib/utils/navigation.ts
```diff
@@ -0,0 +1,13 @@
...
+  navigator: NavigationProp<ReactNavigation.RootParamList>,
+): boolean => {
+  const navigationState = navigator.getState();
+  if (!navigationState) {
+    return false;
+  }
+  const length = navigationState.routes.length;
+  const lastRoute = navigationState.routes[length - 1];
+  return lastRoute.name === 'InProgressActivity';
+};
```
BamMironov, line 13: Please move the hook to either the activity entity as it is related to business logic

BamMironov (changes requested):

src/shared/lib/utils/navigation.ts
anq83: I've moved in to shared as separated function because it also used from 2 places

src/entities/activity/lib/services/AnswersUploadService.ts
anq83: Here we check regular files, not logs

BamMironov (approved):

PR #361, opened 2023-09-05
fix(survey_items): nickname display in variable replacer

DimitrySakalouski:
https://mindlogger.atlassian.net/browse/M2-2943

### DIFF
--- a/src/entities/applet/model/mappers.ts
+++ b/src/entities/applet/model/mappers.ts
@@ -6,6 +6,7 @@ import {
   AnalyticsAnswerDto,
   AnalyticsResponseType,
   AppletDetailsDto,
+  AppletDetailsResponse,
   AppletDto,
   CompletedEntityDto,
   OptionsDto,
@@ -302,4 +303,10 @@ function getItemResponses(
         },
       ];
   }
+}
+
+export function mapDtoToRespondentMeta(
+  response?: AppletDetailsResponse,
+): string {
+  return response?.respondentMeta.nickname ?? '';
 }
--- a/src/entities/applet/model/mappers.ts
+++ b/src/entities/applet/model/mappers.ts
@@ -308,5 +308,5 @@ function getItemResponses(
 export function mapDtoToRespondentMeta(
   response?: AppletDetailsResponse,
 ): string {
-  return response?.respondentMeta.nickname ?? '';
+  return response?.respondentMeta?.nickname ?? '';
 }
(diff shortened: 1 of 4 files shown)


src/features/pass-survey/lib/hooks/useTextVariablesReplacer.ts
```diff
@@ -10,21 +10,28 @@ const useTextVariablesReplacer = ({
...
   items: PipelineItem[] | undefined;
   answers: Answers | undefined;
   activityId: string;
+  appletId: string;
 }) => {
   const completedEntities = useSelector(
     AppletModel.selectors.selectCompletedEntities,
   );
+  const respondentNickname = useAppletDetailsQuery(appletId, {
+    select: r => r.data.respondentMeta.nickname,
```
BamMironov, line 24: Please add the mapper for applet

src/features/pass-survey/lib/hooks/useTextVariablesReplacer.ts
```diff
@@ -10,21 +10,28 @@ const useTextVariablesReplacer = ({
...
 }: {
   items: PipelineItem[] | undefined;
   answers: Answers | undefined;
   activityId: string;
+  appletId: string;
 }) => {
   const completedEntities = useSelector(
     AppletModel.selectors.selectCompletedEntities,
   );
+  const respondentNickname = useAppletDetailsQuery(appletId, {
```
BamMironov, line 23: I believe that respondentNickname should be string already. not object

BamMironov: const { data: respondentNickname } =  useAppletDetailsQuery(...

src/shared/api/services/appletsService.ts
```diff
@@ -76,7 +80,10 @@ type AppletDetailsRequest = {
   appletId: string;
 };
 
-export type AppletDetailsResponse = SuccessfulResponse<AppletDetailsDto>;
+export type AppletDetailsResponse = {
+  result: AppletDetailsDto;
+  respondentMeta: AppletRespondentMetaDto;
```
BamMironov, line 85: What if there is no nickname for the respondent? respondentMeta would be null or nickname would be null or empty string?

BamMironov (changes requested):

src/entities/applet/model/mappers.ts
```diff
@@ -303,3 +304,9 @@ function getItemResponses(
       ];
   }
 }
+
+export function mapDtoToRespondentMeta(
+  response?: AppletDetailsResponse,
+): string {
+  return response?.respondentMeta.nickname ?? '';
```
BamMironov, line 311: Please add another "?" sign before the nickname property
```
response?.respondentMeta?.nickname ?? ''
```

BamMironov (approved):

PR #392, opened 2023-09-21
fix(about_screen): added description to activity analytics card

DimitrySakalouski:
https://mindlogger.atlassian.net/browse/M2-3130


src/entities/applet/lib/types.ts
```diff
@@ -68,7 +68,7 @@ export type ItemResponses = {
 export type ActivityResponses = {
   id: string;
   name: string;
-  description: string | null;
+  description: string | { en: string } | null;
```
BamMironov, line 71: Please come up with a different approach. The description should be provided language agnostic

BamMironov (changes requested):
Please come up with a different solution

PR #405, opened 2023-10-02
test: added ids for testing

DimitrySakalouski:


src/shared/ui/survey/CheckBox/CheckBox.item.tsx
```diff
@@ -90,6 +90,7 @@ const CheckBoxItem: FC<Props> = ({
 
       <CheckBox
         style={styles.checkbox}
+        data-test={`checkbox-${value}`}
```
anq83, line 93: Isn't value true or false here? If so, it cannot be part of test attribute

src/shared/ui/survey/RadioActivityItem/RadioActivityItem.tsx
```diff
@@ -47,6 +47,7 @@ const RadioActivityItem: FC<RadioActivityItemProps> = ({
         value={initialValue?.id ?? ''}
         onValueChange={onValueChange}
         name="radio"
+        data-test="radio-item-group"
       >
         {optionsList.map(option => (
           <Box my="$1" key={option.id} onPress={() => onValueChange(option.id)}>
```
anq83, line 53: Shouldn't we add test attribute into each option?

src/shared/ui/survey/RadioActivityItem/RadioItem.tsx
```diff
@@ -83,20 +86,26 @@ const RadioItem: FC<RadioLabelProps> = ({
         <Box width="10%">
           <CachedImage
             resizeMode="contain"
+            data-test={`radio-option-image-${value}`}
```
anq83, line 89: What is the value here? True / false or text ?

src/widgets/activity-group/ui/ActivityGroup.tsx
```diff
@@ -78,9 +78,18 @@ const ActivityGroup: FC<Props> = ({ appletId, group, ...styledProps }) => {
   };
 
   return (
-    <Box {...styledProps}>
+    <Box
+      data-test={`activity-group-${group.type}-${group.name}`}
```
anq83, line 82: group.name is key for i18, no need to use it, group.type is enough

anq83 (changes requested):

src/shared/ui/survey/RadioActivityItem/RadioItem.tsx
DimitrySakalouski: value is number - unique for each option in radio group

anq83 (approved):

PR #411, opened 2023-10-10
test(applets): added unit tests for refresh logic

DimitrySakalouski:


src/entities/applet/model/services/RefreshService.test.js
```diff
@@ -0,0 +1,133 @@
...
+jest.mock('@app/shared/lib', () => ({
+  getActivityDetailsKey: jest.fn(() => MOCK_ACTIVITY_KEY),
+  collectActivityDetailsImageUrls: jest.fn(() => []),
+  collectAppletDetailsImageUrls: jest.fn(() => []),
+  getAppletDetailsKey: jest.fn(() => MOCK_APPLET_KEY),
+  getEventsKey: jest.fn(() => MOCK_EVENTS_KEY),
+  getAppletsKey: jest.fn(() => MOCK_APPLETS_KEY),
+  isAppOnline: jest.fn(() => Promise.resolve(false)),
+  onNetworkUnavailable: jest.fn(),
+}));
```
BamMironov, line 32: I get why `isAppOnline` and `onNetworkUnavailable` are mocked but the rest of them don't have to be mocked as they return simple JS readable data.

BamMironov (changes requested):

src/entities/applet/model/services/RefreshService.test.js
```diff
@@ -0,0 +1,133 @@
+import { QueryClient } from '@tanstack/react-query';
```
anq83, line 1: Refresh service will be significally changed after merge PR:
https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/424

And with high probability it will be changed again after folowing task:
POC: Optimization based on appletVersion

So, maybe it worth to disable these tests and refrite/fix them afterall
@BamMironov  what do you think?

anq83 (changes requested):

anq83:
Checked the PR, it's not actual as RefreshService has been re-written to support aggregated (getting multiple details) endpoints.
It should be closed an not actual, without merge. @BamMironov

PR #421, opened 2023-10-16
M2-3798: [Mobile App] Application crashes if all items are hidden in …

anq83:
https://mindlogger.atlassian.net/browse/M2-3798

### DIFF
--- a/src/entities/applet/lib/alerts.ts
+++ b/src/entities/applet/lib/alerts.ts
@@ -57,6 +57,6 @@ export function onActivityContainsAllItemsHidden(entityName: string) {
   Alert.alert('', i18n.t('activity:activity_all_items_hidden', { entityName }));
 }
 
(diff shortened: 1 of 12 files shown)


src/entities/applet/lib/alerts.ts
```diff
@@ -34,3 +34,11 @@ export function onMediaReferencesFound() {
     i18n.t('media:media_found_body'),
   );
 }
+
+export function onActivityContainsAllItemsHidden(entityName: string) {
+  Alert.alert('', i18n.t('activity:activity_all_items_hidden', { entityName }));
+}
+
+export function onAnActivityContainsAllItemsHidden(entityName: string) {
```
BamMironov, line 42: Can we name it onFlowContainsAllItemsHidden?

BamMironov (changes requested):

BamMironov (approved):

PR #423, opened 2023-10-17
[WIP] feat(user_account): added user account deletion feature

ghost:


src/shared/lib/alerts/deleteAccountAlert.ts
```diff
@@ -0,0 +1,38 @@
+import { Alert } from 'react-native';
+
+export function onDeleteAccount(onConfirm: () => void) {
+  return new Promise(resolve => {
```
BamMironov, line 4: No need to create or use Promises here

src/shared/lib/alerts/deleteAccountAlert.ts
```diff
@@ -0,0 +1,38 @@
...
+          text: 'Yes, delete',
+          onPress: onConfirm,
+        },
+      ],
+    );
+  });
+}
+
+export function onDeleteAccountConfirmed() {
+  return new Promise(resolve => {
```
BamMironov, line 23: And here also

src/shared/lib/alerts/deleteAccountAlert.ts
```diff
@@ -0,0 +1,38 @@
...
+    Alert.alert(
+      'Account Deletion Request Has Been Sent',
+      `Account Deletion takes up to 24 hours.\n\n
+      During this time, you will be logged out, and access to your account will be blocked.\n\n
+      If you change your mind please contact us
```
BamMironov, line 28: Don't forget to specify translation keys throughout the solution

BamMironov (changes requested):

PR #426, opened 2023-10-17
fix(activities): skip action

DimitrySakalouski:
https://mindlogger.atlassian.net/browse/M2-3832

### DIFF
--- a/src/features/pass-survey/ui/ActivityStepper.tsx
+++ b/src/features/pass-survey/ui/ActivityStepper.tsx
@@ -116,6 +116,12 @@ function ActivityStepper({
     if (!isForced) {
       trackUserAction(userActionCreator.next());
     }
+
+    const currentStepAnswer = activityStorageRecord?.answers[currentStep];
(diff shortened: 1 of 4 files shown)


src/features/pass-survey/ui/ActivityStepper.tsx
```diff
@@ -116,6 +116,12 @@ function ActivityStepper({
     if (!isForced) {
       trackUserAction(userActionCreator.next());
     }
+
+    const currentStepAnswer = activityStorageRecord?.answers[currentStep];
+
+    if (!currentStepAnswer) {
+      trackUserAction(userActionCreator.skip());
+    }
```
BamMironov, line 124: I would also move 120-124 to the "else" block. That's in the case when the user made some change to an activity item but he then was forcibly moved to the next item due to the timer

BamMironov (changes requested):

BamMironov (approved):

PR #446, opened 2023-10-27
Fixed M2-3914 issue with RefreshControl on ios 

vmkhitaryanscn:
Fixed [M2-3914](https://mindlogger.atlassian.net/jira/software/c/projects/M2/issues/M2-3914) issue with RefreshControl on ios  when opening app from background.

https://github.com/facebook/react-native/issues/26801


src/features/applets-refresh/model/hooks/useRefreshResumeOnAppState.ts
```diff
@@ -0,0 +1,38 @@
...
+      if (
+        appState.current.match(/inactive|background/) &&
+        nextAppState === 'active'
+      ) {
+        if (isRefreshing) {
+          setIsRefreshing(false);
+
+          setTimeout(() => {
+            setIsRefreshing(true);
+          }, 350);
```
BamMironov, line 27: Please elaborate on why we need to wrap it in setTimeout and assign the exact delay of 350 ms

BamMironov (changes requested):

src/features/applets-refresh/model/hooks/useRefreshResumeOnAppState.ts
BamMironov: I am wondering if we can utilize [flushSync](https://react.dev/reference/react-dom/flushSync). Could you please give it a try?

src/features/applets-refresh/model/hooks/useRefreshResumeOnAppState.ts
```diff
@@ -0,0 +1,38 @@
...
+        }
+      }
+
+      appState.current = nextAppState;
+    });
+
+    return subscription.remove;
+  }, [isRefreshing, setIsRefreshing]);
```
BamMironov, line 35: Also, I just realized we have `useOnForeground` that you can use instead

BamMironov (changes requested):

src/features/applets-refresh/model/hooks/useRefreshResumeOnAppState.ts
BamMironov: Ah, gotcha :(

BamMironov (approved):

PR #457, opened 2023-11-02
Fixed M2-3979

vmkhitaryanscn:
Fixed M2-3979, checking if the screen is still focused before navigating to InProgressActivity screen

[M2-3979](https://mindlogger.atlassian.net/browse/M2-3979)


src/widgets/activity-group/ui/ActivitySectionList.tsx
```diff
@@ -48,6 +50,10 @@ function ActivitySectionList({ appletId, groups }: Props) {
     entityType: EntityType,
     eventId: string,
   ) {
+    if (!isFocused) {
```
anq83, line 53: It looks weird. We're changing progress state, and make all other multiple checks, but do not do navigation. Why?

anq83 (changes requested):

BamMironov (approved):

PR #458, opened 2023-11-03
M2-3754 unit tests for shared utils

DimitrySakalouski:
### 📝 Description

🔗 [Jira Ticket M2-3754](https://mindlogger.atlassian.net/browse/M2-3754)

Changes include:

- Added unit tests for shared utils


src/shared/lib/utils/common.ts
```diff
@@ -21,7 +21,7 @@ export function range(n: number): number[] {
   return [...Array(n).keys()];
 }
 
-export const isEmptyObject = (object: any) => {
+export const isObjectNotEmpty = (object: any) => {
```
BamMironov, line 24: Please consider this change:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#object-type

```suggestion
export const isObjectNotEmpty = (object: object) => {
```

BamMironov (changes requested):
Looks good but it would be nice to add a more appropriate type to the `isObjectNotEmpty` function

PR #469, opened 2023-11-13
Fixed M2-4067, checking if slider value is null

vmkhitaryanscn:
Fixed [M2-4067](https://mindlogger.atlassian.net/browse/M2-4072), checking if slider value is null, bugfix for continous slider

### DIFF
--- a/src/features/pass-survey/model/AlertsExtractor.ts
+++ b/src/features/pass-survey/model/AlertsExtractor.ts
@@ -94,10 +94,13 @@ class AlertsExtractor {
         sliderAnswer >= alert.minValue &&
         sliderAnswer <= alert.maxValue;
 
-      if (
-        alert.value === sliderAnswer ||
-        (isContinuousSlider && isValueInRange)
-      ) {
+      if (!isContinuousSlider && alert.value === sliderAnswer) {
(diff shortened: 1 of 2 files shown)


src/features/pass-survey/model/AlertsExtractor.ts
```diff
@@ -79,17 +79,24 @@ class AlertsExtractor {
...
+      const isValueInRange =
+        alert.minValue !== null &&
+        alert.maxValue !== null &&
+        sliderAnswer >= alert.minValue &&
+        sliderAnswer <= alert.maxValue;
+
       if (
         alert.value === sliderAnswer ||
-        (sliderAnswer >= alert.minValue && sliderAnswer <= alert.maxValue)
+        (isContinuousSlider && isValueInRange)
```
anq83, line 99: What if slider is not continous? The alerts will never added for such slider type?

anq83 (changes requested):

src/features/pass-survey/model/AlertsExtractor.ts
anq83: ok

anq83 (approved):

PR #473, opened 2023-11-14
feat(drawing): Rebuilt component for the sake of performance

BamMironov:
https://mindlogger.atlassian.net/browse/M2-4100


src/entities/drawer/lib/utils/LineSketcher.ts
```diff
@@ -0,0 +1,92 @@
...
+        path.quadTo(lastPoint.x, lastPoint.y, xMid, yMid);
+      }
+
+      skPaths.push(path);
+    }
+
+    return skPaths;
+  }
+
+  createLine(fromPoint: Point) {
```
anq83, line 53: Why was it named fromPoint whereas it's just point or currentPoint?

src/entities/drawer/lib/utils/LineSketcher.ts
```diff
@@ -0,0 +1,92 @@
+import { MutableRefObject } from 'react';
```
anq83, line 1: Please ensure:
- Bezier working the same as in the previous version
- Dots
- Lines sometimes are'not started drawing, details in the Teams

anq83 (changes requested):

src/entities/drawer/ui/DrawingBoard.tsx
```diff
@@ -1,38 +1,14 @@
...
-  Point,
-  ResponseSerializer,
-  DrawResult,
-  CachedBezierItem,
-  getBezierArray,
-} from '../lib';
+import { DrawLine, ResponseSerializer, DrawResult, LinesTracker } from '../lib';
+import DrawPoint from '../lib/utils/DrawPoint';
 
 const paint = Skia.Paint();
```
anq83, line 13: Please, clean up

src/entities/drawer/ui/DrawingBoard.tsx
```diff
@@ -47,137 +23,49 @@ type Props = {
...
+  const onStrokeChanged = useCallback(
+    (x: number, y: number) => {
+      lineTracker.progressLine(x, y);
+    },
+    [lineTracker],
+  );
 
-    const lines = [...value, newLine];
+  const onStrokeEnd = useCallback(() => {
+    const lines = linesRef.current;
```
anq83, line 68: It seems that coordinates in the DrawLine are in range 0 .. width / height (~300/300). They're not normalized to 100/100.
Is it ok? Was it the same previously?

src/entities/drawer/ui/DrawingBoard.tsx
```diff
@@ -47,137 +23,49 @@ type Props = {
...
```
anq83, line 69: Please check export drawer data in Admin (If you didn't). Just to veryfy that output is similar to the previous version.

anq83 (changes requested):

PR #478, opened 2023-11-16
fix(survey_items): added delay to radio button press

DimitrySakalouski:
https://mindlogger.atlassian.net/browse/M2-4140

### DIFF
--- a/src/features/pass-survey/ui/ActivityItem.tsx
+++ b/src/features/pass-survey/ui/ActivityItem.tsx
@@ -1,4 +1,5 @@
 import { useCallback, useContext, useState } from 'react';
+import { InteractionManager } from 'react-native';
 
 import {
   Box,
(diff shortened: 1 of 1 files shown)


src/features/pass-survey/ui/ActivityItem.tsx
```diff
@@ -90,6 +90,12 @@ function ActivityItem({
     }
   }
 
+  function moveToNextItemWithDelay() {
+    if (!pipelineItem.additionalText?.required) {
+      setTimeout(() => setImmediate(() => next(true)), 200);
```
BamMironov, line 95: Please come up with a different solution
Maybe we can use https://reactnative.dev/docs/interactionmanager#runafterinteractions

BamMironov (changes requested):

BamMironov (approved):

PR #481, opened 2023-11-20
fix(survey_items): flanker image cache

DimitrySakalouski:
https://mindlogger.atlassian.net/browse/M2-4165

### DIFF
--- a/ios/FlankerNativeComponents/FlankerView.swift
+++ b/ios/FlankerNativeComponents/FlankerView.swift
@@ -165,11 +165,7 @@ class FlankerView: UIView {
 
   var isFirstScreen = true
 
-  @objc var dataJson: NSString? {
-    didSet {
-      ParameterGameManager.shared.loadAllImage(dataJson: String(dataJson ?? "")) { _ in }
-    }
-  }
+  @objc var dataJson: NSString?
 
   override init(frame: CGRect) {
     super.init(frame: frame)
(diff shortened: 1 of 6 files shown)


BamMironov (changes requested):

src/entities/flanker/ui/NativeIosFlanker/NativeIosFlanker.tsx
```diff
@@ -31,18 +31,15 @@ const NativeIosFlanker: FC<Props> = props => {
...
-    }, 600);
+    NativeModules.FlankerViewManager.preloadGameImages(configString)
+      .then(() => {
+        NativeModules.FlankerViewManager.setGameParameters(configString);
+        NativeModules.FlankerViewManager.startGame(
+          props.configuration.isFirstPractice,
+          props.configuration.isLastTest,
+        );
+      })
+      .catch((error: string) => Logger.log(error));
```
BamMironov, line 42: Please change the message to `Logger.log('[NativeModules.FlankerViewManager.preloadGameImages] An internal error occurred during caching images or starting the game: \n\n', error)`

ios/FlankerNativeComponents/FlankerView.swift
```diff
@@ -362,13 +358,18 @@ extension FlankerView: GameManagerProtocol {
         rightButton.titleLabel?.textAlignment = .center
       } else if let leftImage = leftImage, let rightImage = rightImage {
         let left = ImageLoader()
-        left.loadImageWithUrl(leftImage)
+        left.downloadImage(url: leftImage).then { image in
+          self.leftButton.setImage(image, for: .normal)
+          self.leftButton.setImage(image, for: .disabled)
+        }
```
BamMironov, line 364: Please handle .catch block just in case

ios/FlankerNativeComponents/FlankerView.swift
```diff
@@ -362,13 +358,18 @@ extension FlankerView: GameManagerProtocol {
...
         let right = ImageLoader()
-        right.loadImageWithUrl(rightImage)
+        right.downloadImage(url: rightImage).then { image in
+          self.rightButton.setImage(image, for: .normal)
+          self.rightButton.setImage(image, for: .disabled)
+        }
```
BamMironov, line 369: Here also

ios/FlankerNativeComponents/FlankerView.swift
```diff
@@ -387,7 +388,7 @@ extension FlankerView: GameManagerProtocol {
         leftButton.titleLabel?.textAlignment = .center
       } else if let leftImage = leftImage {
         let left = ImageLoader()
-        left.loadImageWithUrl(leftImage)
+        let _ = left.downloadImage(url: leftImage)
```
BamMironov, line 391: I am not sure what it does

BamMironov (changes requested):

ios/FlankerNativeComponents/FlankerView.swift
BamMironov: I think we should throw error here in order to catch the error on JS thread

ios/FlankerNativeComponents/FlankerView.swift
BamMironov: I think we should throw error here in order to catch the error on JS thread

BamMironov (changes requested):

BamMironov (approved):

PR #485, opened 2023-11-22
fix(hotfix): sign up terms validation

DimitrySakalouski:


BamMironov:
Fixes
https://mindlogger.atlassian.net/browse/M2-4212

PR #493, opened 2023-11-24
refactor(3110): change test id attribute

DimitrySakalouski:

### DIFF
--- a/src/shared/ui/RoundLogo.tsx
+++ b/src/shared/ui/RoundLogo.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable react-native/no-inline-styles */
 import React, { FC } from 'react';
-import { StyleSheet, ImageStyle } from 'react-native';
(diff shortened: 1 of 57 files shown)


src/shared/ui/RoundLogo.tsx
```diff
@@ -25,7 +25,12 @@ const getImageUrl = (url: string): string => {
   return url;
 };
 
-const RoundLogo: FC<Props> = ({ imageUri, imageStyle, letter, size = 64 }) => {
+const RoundLogo: FC<Props & AccessibilityProps> = ({
```
BamMironov, line 28: AccessibilityProps have been added but not used. Maybe, we should remove it?

BamMironov (changes requested):

BamMironov (approved):

BamMironov (approved):

PR #513, opened 2023-12-01
test: added and changed required test attributes

ghost:


src/shared/ui/Stepper/NextButton.tsx
```diff
@@ -9,9 +10,10 @@ import { HandlersContext } from './contexts';
 
 type Props = PropsWithChildren<{
   isIcon?: boolean;
+  type: StepperNextButtonType | null;
```
BamMironov, line 13: Can we add accessibilityLabel property instead?

src/features/pass-survey/ui/ActivityStepper.tsx
```diff
@@ -105,6 +106,19 @@ function ActivityStepper({
 
   const nextButtonText = getNextButtonText();
 
+  const getNextButtonType = (text: string): StepperNextButtonType | null => {
```
BamMironov, line 109: Let's return here accessibilityLabel instead of a type.

src/features/pass-survey/lib/types/stepperNextButtonType.ts
```diff
@@ -0,0 +1,5 @@
+export enum StepperNextButtonType {
+  DONE = 'done',
+  SKIP = 'skip',
+  NEXT = 'next',
+}
```
BamMironov, line 5: I don't think we need it

BamMironov (changes requested):

PR #529, opened 2023-12-08
Fixed 4348, ignoring timezone when converting DatePicker value to Date

vmkhitaryanscn:
Fixed [4348](https://mindlogger.atlassian.net/jira/software/c/projects/M2/issues/M2-4348), ignoring timezone when converting DatePicker value to Date

### DIFF
--- a/src/shared/lib/utils/dateTime.ts
+++ b/src/shared/lib/utils/dateTime.ts
@@ -205,5 +205,9 @@ export const getDateFromString = (dateString: string) => {
   // converts yyyy-mm-dd string to Date , ignoring timeZone
   const [year, month, day] = dateString.split('-');
 
+  if (!year || !month || !day) {
+    throw new Error('[getDateFromString] dateString format is not valid');
+  }
+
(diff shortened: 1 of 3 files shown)


src/shared/lib/utils/dateTime.ts
```diff
@@ -200,3 +200,10 @@ export const getMonthAgoDate = () => formatToDtoDate(subMonths(new Date(), 1));
 
 export const buildDateTimeFromDto = (yyyymmdd: string, hhmmss: string) =>
   new Date(`${yyyymmdd} ${hhmmss}`);
+
+export const getDateFromString = (dateString: string) => {
+  // converts yyyy-mm-dd string to Date , ignoring timeZone
+  const [year, month, day] = dateString.split('-');
```
BamMironov, line 206: It's worth adding an implicit check if dateString corresponds to the format you specified here.
To do that, you could check if all these variables have values. If not, throw Error

BamMironov (changes requested):

BamMironov (approved):

PR #550, opened 2023-12-19
M2-4436 add timeout for upload

anq83:
https://mindlogger.atlassian.net/browse/M2-4436

We'll remove WIP when it's possible to test it. Currently I do not catch any hangs and only getting 413 using iOS Simulator 14.
Whereas ~ a week ago - it was easy to get hang.

No sure is the solution correct. Probably - the "promise" which is responsible for delivering response is being removed is such cases.

### DIFF
--- a/src/entities/activity/lib/services/AnswersUploadService.ts
+++ b/src/entities/activity/lib/services/AnswersUploadService.ts
@@ -118,13 +118,17 @@ class AnswersUploadService implements IAnswersUploadService {
   ): Promise<string> {
     const localFileExists = await FileSystem.exists(mediaFile.uri);
 
-    const logFileInfo = `(${mediaFile.type}, from answer #${logAnswerIndex})`;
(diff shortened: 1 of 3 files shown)


src/entities/activity/lib/services/AnswersUploadService.ts
```diff
@@ -118,14 +118,18 @@ class AnswersUploadService implements IAnswersUploadService {
...
 
     if (!localFileExists) {
       throw new Error(
         `[UploadAnswersService.processFileUpload]: Local file ${logFileInfo} does not exist`,
       );
     }
 
+    const fileStat = await FileSystem.stat(mediaFile.uri);
+
+    logFileInfo += `, with size = ${fileStat.size} bytes`;
```
BamMironov, line 131: Could we move the declaration of the `logFileInfo` variable here so that we pass the value once?

BamMironov (changes requested):

PR #552, opened 2023-12-19
M2-4428 not showing loader on android, when audio track is resumed playing

vmkhitaryanscn:
[M2-4428](https://mindlogger.atlassian.net/browse/M2-4428) not showing loader on android, when audio track is resumed playing

[M2-4428]: https://mindlogger.atlassian.net/browse/M2-4428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/shared/lib/hooks/useAudioPlayer.ts
+++ b/src/shared/lib/hooks/useAudioPlayer.ts
@@ -7,10 +7,18 @@ import { IS_ANDROID, wait } from '@shared/lib';
 
 const SUBSCRIPTION_DURATION = 500;
 
+enum LoadingStates {
+  'initial' = 0,
+  'loading' = 1,
+  'loaded' = 2,
+}
+
 const useAudioPlayer = () => {
   const audioRecorderPlayer = useRef(new AudioRecorderPlayer());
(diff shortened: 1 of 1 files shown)


src/shared/lib/hooks/useAudioPlayer.ts
```diff
@@ -10,7 +10,7 @@ const SUBSCRIPTION_DURATION = 500;
 const useAudioPlayer = () => {
   const audioRecorderPlayer = useRef(new AudioRecorderPlayer());
   const [isPlaying, setIsPlaying] = useState(false);
-  const [isLoading, setIsLoading] = useState(false);
+  const [isLoading, setIsLoading] = useState<null | boolean>(null);
```
BamMironov, line 13: What if we pass `true` instead as an initial value? This way, you will not have to do `setIsLoading(true)` at the start of the `play` function at all.

BamMironov (changes requested):

BamMironov (approved):

PR #554, opened 2023-12-20
M2-4462: [Mobile app] Extra Notifications for recurring cross-day eve…

anq83:
[…nt are scheduledhttps://mindlogger.atlassian.net/browse/M2-4462](https://mindlogger.atlassian.net/browse/M2-4462)

### DIFF
--- a/src/entities/notification/lib/types/notificationBuilder.ts
+++ b/src/entities/notification/lib/types/notificationBuilder.ts
@@ -129,7 +129,7 @@ export type NotificationBuilderInput = {
 export type EventNotificationDescribers = {
   eventId: string;
   eventName: string;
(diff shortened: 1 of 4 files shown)


src/entities/notification/lib/types/notificationBuilder.ts
```diff
@@ -129,6 +129,7 @@ export type NotificationBuilderInput = {
 export type EventNotificationDescribers = {
   eventId: string;
   eventName: string;
+  eventObject: ScheduleEvent;
```
BamMironov, line 132: Please come up with a different name, e.g. scheduleEvent

BamMironov (changes requested):

BamMironov (approved):

PR #560, opened 2023-12-26
feat(survey_items): added enhancements for auto test attributes

DimitrySakalouski:

### DIFF
--- a/src/entities/activity/ui/ActivityCard.tsx
+++ b/src/entities/activity/ui/ActivityCard.tsx
@@ -1,5 +1,5 @@
-import { FC } from 'react';
-import { AccessibilityProps, TouchableOpacity } from 'react-native';
+import { FC, useMemo } from 'react';
+import { TouchableOpacity } from 'react-native';
 
 import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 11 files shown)


src/entities/activity/ui/ActivityCard.tsx
```diff
@@ -23,16 +23,22 @@ type Props = {
...
+  const accessibilityLabel = useMemo(
+    () =>
+      `activity-${
+        activity.isInActivityFlow &&
+        activity.activityFlowDetails!.showActivityFlowBadge
+          ? 'flow-'
+          : ''
+      }${activity.name}`,
+    [activity],
+  );
```
BamMironov, line 40: There is no need to use useMemo here

src/entities/activity/ui/ActivityList.tsx
```diff
@@ -16,7 +16,6 @@ const ActivityList: FC<Props> = ({ activities, onCardPress }) => {
       {activities.map(x => (
         <Box key={x.eventId}>
           <ActivityCard
-            accessibilityLabel={`activity-${x.name}`}
```
BamMironov, line 19: isn't that a mistake that we are deleting this?

BamMironov (changes requested):

src/entities/activity/ui/ActivityList.tsx
anq83: no, the attribute has been moved into ActivityCard

BamMironov (approved):

PR #563, opened 2023-12-29
Unit tests for collectImageUrls functions

vmkhitaryanscn:
https://mindlogger.atlassian.net/browse/M2-4530

### DIFF
--- a/src/shared/lib/services/collectImageUrls.test.ts
+++ b/src/shared/lib/services/collectImageUrls.test.ts
@@ -527,7 +527,7 @@ describe('Image URL Collection Functions', () => {
     streamEnabled: false,
   };
 
-  it('collectAppletRecordImageUrls should collect image URLs from AppletDto', () => {
+  it('should collect image URLs from AppletRecord', () => {
     const result = collectAppletRecordImageUrls(sampleApplet);
     expect(result).toEqual([
       'https://media-dev.cmiml.net/mindlogger/content/2923f4a9-20ef-4995-a340-251d96ff3082/0e06365a-acc7-4db9-b7fc-8a0017a2354f/%D0%A8%D0%AC%D0%A4%D0%9F%D0%A3%206.png',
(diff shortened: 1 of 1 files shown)


src/shared/lib/services/collectImageUrls.test.ts
```diff
@@ -0,0 +1,578 @@
...
+      'https://media-dev.cmiml.net/mindlogger/content/2923f4a9-20ef-4995-a340-251d96ff3082/0e06365a-acc7-4db9-b7fc-8a0017a2354f/%D0%A8%D0%AC%D0%A4%D0%9F%D0%A3%206.png',
+      'https://media-dev.cmiml.net/mindlogger/content/2923f4a9-20ef-4995-a340-251d96ff3082/68dff67f-47af-4ba9-adec-9440077b0e3d/IMAGE%202.png',
+      'https://media-dev.cmiml.net/mindlogger/content/some-logo.png',
+      'https://media-dev.cmiml.net/mindlogger/content/some-backgroundImage.png',
+      'https://media-dev.cmiml.net/mindlogger/content/2923f4a9-20ef-4995-a340-251d96ff3082/c26556ef-ef83-4782-9397-ea017a534e1c/%D0%A8%D0%AC%D0%A4%D0%9F%D0%A3%206.png',
+      'https://media-dev.cmiml.net/mindlogger/content/2923f4a9-20ef-4995-a340-251d96ff3082/c22ba829-ad08-4a7c-bbe6-9fd5a719f7e3/IMAGE%202.png',
+    ]);
+  });
+
+  it('collectActivityDetailsImageUrls should collect image URLs from ActivityDto', () => {
```
BamMironov, line 552: Please remove "collectActivityDetailsImageUrls" word from string parameter of "it" functions.
When we call "it" function we should stick to the code style so that when we read the code line we build a sentence using the following pattern: it("should do something") / it("returns something").
Example:
image

BamMironov: Please apply the same change to every new test case

BamMironov (changes requested):

BamMironov (approved):

PR #606, opened 2024-02-01
M2-4824: include password requirements hint

jgdev:
Solves ticket: [M2-4824](https://mindlogger.atlassian.net/browse/M2-4824)

image

[M2-4824]: https://mindlogger.atlassian.net/browse/M2-4824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/shared/ui/InputIcon.tsx
```diff
@@ -0,0 +1,32 @@
...
+    right: 10,
+  },
+});
+
+export function InputIcon<T = any>({ input, icon }: InputIconProps<T>) {
+  return (
+    <View style={styles.inputContainer}>
+      <View style={styles.input}>{input}</View>
+      <View style={styles.icon}>{icon}</View>
+    </View>
```
farmerpaul, line 30: Just a few little things to mention:

- I think we should take advantage of Tamagui components whenever possible, as it will reduce the amount of styles we have to write as well as put us in a better position for eventually unifying the UI component library across FE apps. I haven't seen this made into a best practice document yet, but I'd like to push for it.
- Generally, I'd avoid absolutely positioning an icon contained inside an inputs, and use natural flexbox layout instead, else we tend to run into situations like this… 😏
  Screenshot 2024-02-01 at 9 02 17 PM
- However, the solution I had in mind:
  ```suggestion
    <XStack alignItems="center">
      <Box flex={1}>{input}</Box>
      <Box px={10}>{icon}</Box>
    </XStack>
  ```
  doesn't _quite_ solve it because the bottom border focus effect is being applied only inside the `input` and not _this_ higher level component. 😔 So a perfect solution would probably require integrating the right icon as an optional prop on the `InputField` component, and placing the above `XStack` in there.

src/shared/ui/form/InputPassword.tsx
```diff
@@ -0,0 +1,101 @@
...
+                secureTextEntry={isPasswordHidden}
+                {...props}
+                isInvalid={isInvalid || !!error}
+              />
+            }
+            icon={
+              <TouchableWithoutFeedback
+                onPress={() => setPasswordHidden(!isPasswordHidden)}
+              >
+                <ShowPasswordIcon size={24} color={mode} />
```
farmerpaul, line 86: The icon is rendering in black on iOS, but I think it's meant to be white (it works on Android):

Screenshot 2024-02-01 at 9 19 13 PM

src/shared/ui/icons/index.tsx
```diff
@@ -119,3 +120,19 @@ export const ChevronDownIcon: FC<IconProps> = props => (
...
+  <FontAwesome name="eye-slash" {...props} />
+);
+
+export const FeatherCrossIcon: FC<IconProps> = props => (
+  <Feather name="x" {...props} />
+);
+
+export const FeatherCheckIcon: FC<IconProps> = props => (
+  <Feather name="check" {...props} />
+);
```
farmerpaul, line 138: The "Feather" font seems to be causing errors on iOS; I guess you have to add the font there, too?

src/shared/ui/form/tests/InputPassword.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+  it('should display validation when the password have 8 characters', () => {
+    expect(
+      getPasswordRequirements(
+        createTest({
+          defaultValues: {
+            password: 'testpassword',
+          },
+        }).instance,
+      ).minCharacters.icon.props.color,
+    ).toBe(colors.white);
```
farmerpaul, line 120: In general, relying on colour comparison to evaluate component behaviour doesn't quite feel ideal to me, for a couple reasons:

- Designs are often evolving, and different colours might be used to describe validity or non-validity states, making these tests possibly fragile
- In general, we should be keeping a11y top of mind whenever possible while writing unit tests, which to me often boils down to "how would a screen reader describe the screen state?" Screen readers generally don't care about the colour of text, but they do pay attention to a11y labels and other a11y props.

By thinking about how you'd navigate and get information about the screen state via, say, TalkBack (on Android) or VoiceOver (on iOS), it encourages us to write UIs that are accessible.

In this case, in the `PasswordRequirement` component, I might give it an a11y prop to indicate what state it is in. Maybe if the requirement is not met, I'd assign it `accessibilityState={{ disabled: true }}`, and then compare against that in my unit test. You could also use an `accessibilityLabel` to explain validation state in more words, but then you may need to add translation strings.

Having said all this, I also know that a11y can be tricky to get right and could take you down some deep rabbit holes, especially as Android and iOS don't perfectly align in behaviour. So I'm totally fine if you want to keep things the way you have them for now! It's still early on the project and a11y might not be a priority for a while yet.

src/shared/ui/form/PasswordRequirements.tsx
```diff
@@ -0,0 +1,121 @@
...
+  return (
+    <StyledView testID={label}>
+      <StyledIconView>
+        <Icon color={iconColor} size={16} />
+      </StyledIconView>
+
+      <Text fontSize={14} marginLeft={8} color={textColor}>
+        {label}
+      </Text>
+    </StyledView>
```
farmerpaul, line 86: Just want to suggest using Tamagui again, as I think a lot of the above `styled` calls could be reduced or eliminated just by using Tamagui components and props. But no big deal for now, just something to keep in mind for later!

farmerpaul (changes requested):
Hey @jgdev, lots of great work here! I tested and it's working fine on Android, but on iOS I'm getting a font error and a minor styling inconsistency, which I've mentioned in comments below.

I also just made a bunch of suggestions and things to keep in mind/discuss later on. 🙂

But great first PR! 🥳

src/shared/ui/form/ErrorMessage.tsx
```diff
@@ -16,21 +16,19 @@ const ErrorMessage: FC<PropsWithChildren<Props>> = ({
...
-          fontSize={12}
-          {...props}
-        >
-          {/* @ts-ignore */}
-          {t(error.message, error.params)}
-        </Text>
-      )}
-    </>
-  );
+  return <>{error?.message && renderError(error?.message)}</>;
```
anq83, line 31: Shouldn't we add !!error?.message  to ensure boolean?

anq83 (dismissed):

src/shared/ui/form/tests/InputPassword.test.tsx
```diff
@@ -0,0 +1,194 @@
+import React from 'react';
+import { TextInput } from 'react-native';
+
+import { FormProvider } from 'react-hook-form';
+import renderer, { act } from 'react-test-renderer';
+
+import TamaguiProvider from '@app/app/ui/AppProvider/TamaguiProvider';
+import SignUpFormSchema from '@app/features/sign-up/validation/SignUpFormSchema';
```
BamMironov, line 8: According to FSD, we must not access the upper layers

src/shared/ui/form/tests/InputPassword.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+import { colors } from '@app/shared/lib/constants/colors';
+import useAppForm from '@app/shared/lib/hooks/useAppForm';
```
BamMironov, line 10: Please use the following patter for imports:

`import { something } from '@app/shared/lib'
`or
`import { something } from '@app/shared/ui'`

src/features/sign-up/validation/SignUpFormSchema.ts
```diff
@@ -1,16 +1,74 @@
...
+        message: upperCase.label,
+      });
+    !noSpaces.isValid &&
+      refinement.addIssue({
+        code: 'custom',
+        message: noSpaces.label,
+      });
+
+    return refinement;
+  }),
```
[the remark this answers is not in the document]
BamMironov: @farmerpaul Validation is not a responsibility of a field but a form

BamMironov (changes requested):

src/features/sign-up/validation/SignUpFormSchema.ts
```diff
@@ -1,16 +1,74 @@
...
+      isValid: password.length > 0 && containsUpperCase,
+      label: 'password_requirements:at_least_capital_letter',
+    },
+    noSpaces: {
+      isValid: password.length > 0 && doesNotIncludeSpaces,
+      label: 'password_requirements:no_blank_spaces',
+    },
+  };
+};
+
```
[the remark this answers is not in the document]
mbanting: Great catch @farmerpaul I've added it to JIRA, [M2-5095](https://mindlogger.atlassian.net/browse/M2-5095)

src/features/sign-up/ui/tests/__snapshots__/SignUpForm.test.tsx.snap
farmerpaul, line 1: Does this snapshot need to be updated? `yarn jest -u`

farmerpaul (approved):
Nice work @jgdev! I understand some of the tests might be failing due to an API breaking change? There's also one snapshot test that's failing, I mentioned it below. But preapproving, pending tests passing! 🚀

src/features/sign-up/ui/SignUpForm.tsx
```diff
@@ -34,14 +47,35 @@ const SignUpForm: FC<Props> = props => {
     onSubmitSuccess: data => {
       executeIfOnline(() => signUp(data));
     },
+    criteriaMode: 'all',
```
BamMironov, line 50: Nice catch!  I completely forgot about this one

src/features/sign-up/ui/tests/__snapshots__/SignUpForm.test.tsx.snap
BamMironov, line 1: Should we use React Native's Virtual DOM elements instead of browser's DOM?
https://jestjs.io/docs/tutorial-react-native#snapshot-test

src/shared/ui/form/ErrorMessage.tsx
```diff
@@ -6,7 +6,7 @@ import { useTranslation } from 'react-i18next';
 import { Text, TextProps } from '@shared/ui';
 
 type Props = {
-  error?: FieldError | { message: string };
+  error?: Partial<FieldError>;
```
BamMironov, line 9: Please double-check if it doesn't break the other error messages that are used in other forms. Especially, when a **backend error** is passed to the component

BamMironov (changes requested):
A few comments but overall, nice job!

src/shared/ui/form/ErrorMessage.tsx
jgdev: Already tested and working 👍🏽

src/features/sign-up/ui/SignUpForm.tsx
```diff
@@ -34,14 +47,35 @@ const SignUpForm: FC<Props> = props => {
...
+    shouldUseNativeValidation: false,
```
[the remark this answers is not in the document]
jgdev: This is because of the `zodResolver` function as you can see [here](https://github.com/ChildMindInstitute/mindlogger-app-refactor/blob/df1a0ea0fc577da2e0756e703d6830405946d540/src/shared/lib/validation/zodResolver.ts#L93), by removing it or setting it to `true` will result in not retrieve all errors at once.

src/features/sign-up/ui/SignUpForm.tsx
```diff
@@ -34,14 +47,35 @@ const SignUpForm: FC<Props> = props => {
...
   });
 
   useFormChanges({
     form,
     watchInputs: ['password'],
-    onInputChange: () => reset(),
+    onInputChange: function () {
+      form.trigger('password');
+    },
```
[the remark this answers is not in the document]
jgdev: Yes, it is not longer necessary using `criteriaMode=all` .

src/features/sign-up/ui/SignUpForm.tsx
```diff
@@ -34,14 +47,35 @@ const SignUpForm: FC<Props> = props => {
...
+    ).flat();
+
+    return [
+      'password_requirements:at_least_characters',
+      'password_requirements:no_blank_spaces',
+    ].map(key => ({
+      label: key,
+      isValid: passwordHasLength && !errors.includes(key),
+    }));
+  }, [passwordHasLength, form.formState.errors?.password?.types]);
```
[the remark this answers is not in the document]
jgdev: Much way better, but I still need to use it as `{isPasswordFocus && <SignUpPasswordRequirements />}` in order to follow the [UX/UI](https://www.figma.com/file/NVP7wLyiN0U80HZI7w6UU0/Bugs-and-Fast-Follows?type=design&node-id=9-6906&mode=design&t=vPL5ZWFEWE7EsYeP-0) and the [task](https://mindlogger.atlassian.net/browse/M2-4824).

src/features/sign-up/ui/SignUpForm.tsx
```diff
@@ -34,14 +47,35 @@ const SignUpForm: FC<Props> = props => {
...
+  const passwordHasLength = !!form.getValues().password.length;
+  const passwordRequirements = useMemo(() => {
+    const errors = Object.values(
+      form.formState.errors?.password?.types || {},
+    ).flat();
+
+    return [
+      'password_requirements:at_least_characters',
+      'password_requirements:no_blank_spaces',
+    ].map(key => ({
```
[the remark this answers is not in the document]
jgdev: Following the [UX/UI](https://www.figma.com/file/NVP7wLyiN0U80HZI7w6UU0/Bugs-and-Fast-Follows?type=design&node-id=9-6906&mode=design&t=vPL5ZWFEWE7EsYeP-0) and the [task](https://mindlogger.atlassian.net/browse/M2-4824), I have the display the requirements as soon as the input is focused, at that time there's no error to display, so all the errors to be checked must be defined here in the initial state to display it as the example:

image

And validate them while typing.

I am not sure what would be the best for this scenario.

src/features/sign-up/ui/tests/__snapshots__/SignUpForm.test.tsx.snap
jgdev: Tests are configured as expected using React Native's Virtual DOM, by other hand tamagui components seems to be rendered as browser DOM elements. 🤔

jgdev: Nevermind, just figure it out, it seems to be a local configuration, snapshots are being generated differently on Github Actions.

src/features/sign-up/ui/SignUpForm.tsx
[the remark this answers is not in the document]
BamMironov: @jgdev  yes, sure!

src/shared/ui/form/PasswordRequirements.tsx
```diff
@@ -0,0 +1,103 @@
+import { View } from 'react-native';
+
+import { styled, Text } from '@tamagui/core';
+import { useTranslation } from 'react-i18next';
+
+import { colors } from '@shared/lib/constants/colors';
```
BamMironov, line 6: Let's put it that way. I promise I'll create a ESLint rule to address the `import` code styles.

```suggestion
import { colors } from '@shared/lib';
```

src/shared/ui/form/InputField.tsx
```diff
@@ -3,25 +3,31 @@ import { ColorValue, TextInputProps } from 'react-native';
 
 import { Controller, useFormContext, useController } from 'react-hook-form';
 
-import { Input } from '@shared/ui';
+import { Box, Input, XStack } from '@shared/ui';
 import { ErrorMessage } from '@shared/ui/form';
 
-type Props = {
+export type InputFieldProps = {
```
BamMironov, line 9: That's actually optional but makes sense to do as we don't want to expose unnecessary types because there are already too many of them. Since it is not used anywhere and most likely will not be used let's not export it.
Also, we usually name it just `Props` (if it's not exported) because, in our opinion, it is easier to locate them when one is looking at the file. Or maybe it is just the code style of choice :)

```suggestion
type Props = {
```

BamMironov (approved):
That looks great!
Thank you @jgdev, nice job

I left a couple of suggestions regarding imports but no worries

src/shared/ui/form/InputField.tsx
jgdev: My bad, I thought I returned it back to how it was before.

PR #622, opened 2024-02-15
M2-4825: Recovery password notification

jgdev:
Ticket: [M2-4825](https://mindlogger.atlassian.net/browse/M2-4825)

### 📝 Description

After the user requests the reset password, should receive a banner/toast notification.

### 📸 Screenshots

Screenshot 2024-02-20 at 7 12 36 PM
Screenshot 2024-02-20 at 7 14 07 PM

[M2-4825]: https://mindlogger.atlassian.net/browse/M2-4825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


android/settings.gradle
```diff
@@ -9,3 +9,6 @@ if (settings.hasProperty("newArchEnabled") && settings.newArchEnabled == "true")
     include(":ReactAndroid:hermes-engine")
     project(":ReactAndroid:hermes-engine").projectDir = file('../node_modules/react-native/ReactAndroid/hermes-engine')
 }
+
+include ':react-native-vector-icons'
+project(':react-native-vector-icons').projectDir = new File(rootProject.projectDir, '../node_modules/react-native-vector-icons/android')
```
BamMironov, line 14: According to the [docs](https://www.npmjs.com/package/react-native-vector-icons), these lines are optional and are only necessary if you intend to utilize the `Icon.getImageSource` function.
But I don't see you used it.
What was the reason for adding those lines?

assets/translations/en.json
```diff
@@ -168,8 +168,9 @@
     "enter_email_mess": "Please enter your email address",
     "invalid_email": "Invalid email address",
     "email_address": "Email address",
-    "email_sent": "Reset email has been sent",
-    "email_send_error": "That email does not exist in the system.",
+    "email_sent_1": "Reset email has been sent to",
+    "email_sent_2": "if the account exists.",
+    "email_send_error": "Unable to send reset email at this time.",
```
BamMironov, line 173: Maybe, we should move it from "forgot_pass_form" into a separate group of keys?

ios/MindloggerMobile/Info.plist
```diff
@@ -67,7 +67,9 @@
 		<string>Foundation.ttf</string>
 		<string>Entypo.ttf</string>
 		<string>MaterialIcons.ttf</string>
+    <string>MaterialCommunityIcons.ttf</string>
 		<string>Feather.ttf</string>
+    <string>Octicons.ttf</string>
```
BamMironov, line 72: Hmm, the format is different though. It seems like it has different tab/spaces

ios/Podfile.lock
```diff
@@ -951,4 +951,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 3fd9f605d2ca061c942789de32928fa0ba430aa8
 
-COCOAPODS: 1.11.3
+COCOAPODS: 1.15.0
```
BamMironov, line 954: Please revert it. 1.15.0 has changes that won't be working with the latest React Native.
Please install 1.11.3 in your local machine

src/app/ui/AppProvider/ToastProvider.tsx
```diff
@@ -3,7 +3,17 @@ import { FC, PropsWithChildren } from 'react';
...
-import { Box, Text } from '@app/shared/ui';
+import { colors } from '@app/shared/lib';
+import {
+  Box,
+  Text,
+  MaterialAlertOctagon,
+  OcticonsCircleCheckFill,
+} from '@app/shared/ui';
+import { useSafeAreaInsets } from 'react-native-safe-area-context';
+import { Platform } from 'react-native';
```
BamMironov, line 14: Did you set up auto-format on save? It seems like the editor should've automatically fixed the import order, but it didn't for some reason.
According to the ESLint rules, react-native should be on top and react-native-safe-area-context should be before react-native-toast-notifications

src/app/ui/AppProvider/BannerProvider.tsx
BamMironov, line 1: Empty file :)

src/app/ui/AppProvider/ToastProvider.tsx
```diff
@@ -3,7 +3,17 @@ import { FC, PropsWithChildren } from 'react';
...
+
+const isIOS = Platform.OS === 'ios';
```
BamMironov, line 16: Please use `IS_IOS` constant. You can find it on "@shared/lib"

src/app/ui/AppProvider/ToastProvider.tsx
```diff
@@ -13,8 +23,10 @@ const ToastButton = styled(Box, {
   padding: 10,
 });
 
-type ToastProps = {
+export type ToastProps = {
```
BamMironov, line 26: Please rename it to `Props` and don't export it unless it is really needed

src/app/ui/AppProvider/ToastProvider.tsx
```diff
@@ -13,8 +23,10 @@ const ToastButton = styled(Box, {
...
   message: string | JSX.Element;
+  icon?: React.ReactNode;
```
BamMironov, line 28: I believe JSX.Element would be a better type here

src/app/ui/AppProvider/index.tsx
BamMironov, line 1: [React Navigation](https://reactnavigation.org/docs/getting-started) already has `SafeAreaProvider` built-in in the `NavigationProvider`. So hooks like `useSafeAreaInsets` already work fine.

src/features/forgot-password/ui/ForgotPasswordForm.tsx
```diff
@@ -28,16 +41,29 @@ const ForgotPasswordForm: FC<Props> = props => {
...
+          <Text fontSize={14}>
+            {t('forgot_pass_form:email_sent_1')}
+            {'\n'}
+          </Text>
+          <Text fontSize={14} fontWeight="800">
+            {form.getValues().email}
+            {'\n'}
+          </Text>
+          <Text fontSize={14}>{t('forgot_pass_form:email_sent_2')}</Text>
+        </Text>,
```
BamMironov, line 55: Let's move it to a separate component, shall we?

src/shared/lib/hooks/useBanner.ts
```diff
@@ -0,0 +1,28 @@
+import { ToastProps } from '@app/app/ui/AppProvider/ToastProvider';
```
BamMironov, line 1: According to [FSD](https://feature-sliced.design/docs/get-started/overview), lower layer are not allowed to access higher layers

BamMironov (changes requested):
Everything is looking good. Left a few comments.
Also, please review my comment here:
https://mindlogger.atlassian.net/browse/M2-4825?focusedCommentId=18524
I would ask you to do that in this PR if you don't mind

BamMironov:
By the way, what does it look like on an iPad/Tablet?

src/app/ui/AppProvider/ToastProvider.tsx
```diff
@@ -27,15 +39,70 @@ const Toast: FC<ToastProps> = ({ message }) => (
...
+};
+
+const Banner: FC<ToastProps> = (props: ToastProps) => {
+  const { top } = useSafeAreaInsets();
+  const { icon, bgColor } = getDefaultBannerProps(props);
+  return (
+    <Box
+      backgroundColor={bgColor}
+      width="100%"
+      py={40}
```
mbanting, line 70: Should be a smaller value. See the corresponding [banner design](https://www.figma.com/file/vFUwmYD8Ya0ShZsJJ2f757/Web-app?type=design&node-id=1535-31902&mode=design) for the web app, which Mike X mentioned we can use as reference for mobile

android/settings.gradle
jgdev: My bad here, I was testing something that wasn't working, thank you.

assets/translations/en.json
```diff
@@ -168,8 +168,9 @@
     "enter_email_mess": "Please enter your email address",
     "invalid_email": "Invalid email address",
     "email_address": "Email address",
-    "email_sent": "Reset email has been sent",
-    "email_send_error": "That email does not exist in the system.",
+    "email_sent_1": "Reset email has been sent to",
+    "email_sent_2": "if the account exists.",
```
farmerpaul, line 172: This is fine, but just some food for thought:

Ideally this would just be 1 translation key using token replacement, which makes translation keys less confusing to translators:
```suggestion
    "email_send_success": "Reset email has been sent to\n<1>{{email}}</1>\nif the account exists.",
```

Then you'd have something like this in your code:

```tsx
        <Text fontSize={14}>
          <Trans i18nKey="forgot_pass_form:email_send_success">
            Reset email has been sent to
            <Text fontSize={14} fontWeight="800">
              <>{{ email: form.getValues().email }}</>
            </Text>
            if the account exists.
          </Trans>
        </Text>
```

I haven't tested it, there might be some minor corrections needed above.

This technique for replacing tokens inside rich text is often used in the admin panel (see for examples).

src/app/ui/AppProvider/ToastProvider.tsx
farmerpaul: +1

If you use VS Code, then you should have the ESLint extension enabled, and configured something like this in your VS Code settings:
```json
  "eslint.format.enable": true,
  "editor.formatOnSave": true,
```

Then this code should be flagged with eslint errors:

Screenshot 2024-02-20 at 6 29 43 PM


and automatically fixed to this save:

Screenshot 2024-02-20 at 6 29 49 PM

src/app/ui/AppProvider/ToastProvider.tsx
```diff
@@ -13,8 +23,10 @@ const ToastButton = styled(Box, {
   padding: 10,
 });
 
-type ToastProps = {
+export type ToastProps = {
   message: string | JSX.Element;
+  icon?: React.ReactNode;
+  type?: string;
```
farmerpaul, line 29: I'm not sure a freeform `string` type is appropriate here. I would type it as below:

```suggestion
  type?: 'success' | 'warning' | 'danger' | 'info';
```

since the Banner component is prescriptively styled in Figma. This also improve developer experience as you can just rely on built-in autocomplete rather than have to open you component code to figure out what options are allowed.

(It's also fine if you want to limit your implementation to just `success` and `danger` for the time being, since that's all you need for your feature, but I'd still be prescriptive in your typing.)

src/app/ui/AppProvider/ToastProvider.tsx
```diff
@@ -27,15 +39,70 @@ const Toast: FC<ToastProps> = ({ message }) => (
...
-    offsetBottom={20}
-    renderToast={({ message }) => {
-      return <Toast message={message} />;
-    }}
-  >
-    {children}
-  </RNTNToastProvider>
-);
+const getDefaultBannerProps = (props: ToastProps) => {
+  let type = props.type?.replace('banner', '').toLowerCase(),
```
farmerpaul, line 43: I'm curious what is the purpose of this string replacement, can you explain?

I would just go with a default value of `success` if `type` is not supplied.

src/app/ui/AppProvider/index.tsx
jgdev: @BamMironov As it's being used inside `ToastProvider` seems not to have access to it.

image

A solution could be instantiate it inside `ToastProvider` itself.

src/app/ui/AppProvider/ToastProvider.tsx
```diff
@@ -27,15 +39,70 @@ const Toast: FC<ToastProps> = ({ message }) => (
...
+
+const Banner: FC<ToastProps> = (props: ToastProps) => {
+  const { top } = useSafeAreaInsets();
+  const { icon, bgColor } = getDefaultBannerProps(props);
+  return (
+    <Box
+      backgroundColor={bgColor}
+      width="100%"
+      py={40}
+      pt={isIOS ? top + 40 : 40}
```
farmerpaul, line 71: In general, I wouldn't assume that just because a device is not iOS, that it doesn't have a top safe area inset that should be taken into account. Do you know this for certain? I would expect that `useSafeAreaInsets()` returns an appropriate value for `top` for most Android devices, so I wouldn't expect to have to add another OS check here – but please correct me if I'm wrong, I haven't tested. 🙂

src/app/ui/AppProvider/index.tsx
jgdev: I've been trying different approaches, no one seems to work.

farmerpaul (commented):
Hi @jgdev, I left a few initial comments, but I don't have time to go through the whole PR right now. I'll try to get through the rest tomorrow. 🙂

assets/translations/en.json
jgdev: It works this way:

```jsx
// Component
<Text>
  <Trans
    i18nKey="forgot_pass_form:email_send_success"
    components={{
      strong: <Text fontSize={14} fontWeight="800" />,
    }}
    values={{ email: props.email }}
  />
</Text>
```

```js
// Translation
{
  "email_send_success": "Reset email has been sent to <strong>{{email}}</strong> if the account exists.",
}
```

just amazing !! thanks @farmerpaul

src/app/ui/AppProvider/ToastProvider.tsx
jgdev: Also adding `warning` and `info` impl ;)

assets/translations/en.json
jgdev: as a suggestion from @farmerpaul I'm thinking to keep it as 
```
{
  "email_send_success": "Reset email has been sent to <strong>{{email}}</strong> if the account exists.",
}
```

src/app/ui/AppProvider/index.tsx
farmerpaul: @BamMironov in the current React Navigation docs they actually do explicitly wrap `<NavigationContainer>` with `<SafeAreaProvider>` in [their example code](https://arc.net/l/quote/vyqsqoec), so perhaps it is still required? (On my last project, we also explicitly included `<SafeAreaProvider>` within our app provider.)

src/app/ui/AppProvider/ToastProvider.tsx
jgdev: As we discussed on Slack:

The banner is actually using the top safe area insets, this will place the banner normally, right down the status bar both ios/android as normally should be.

The mean of using `mt={IS_IOS ? -top : -12}` will put the banner at the top like if the status bar doesn't exists, but at the same time, Using the padding `pt={IS_IOS ? top + 12 : 12}`  will place the content on the right place bellow the status bar.

The intention is for IOS to remove the blue background behind the notch and use the same as the banner background, to have a smooth design.

image

In android this is not needed since the status bar always remain black or with an accent color.

jgdev:
All comments solve by now 👍🏽

BamMironov (approved):
Looks good! Just one comment from my side regarding translation keys. I think it's not a big deal to fix it

PR #627, opened 2024-02-19
M2-5267: [Mobile] Report summary screen: add aqa test-attributes

anq83:
### 📝 Description

🔗 [Jira Ticket M2-5267](https://mindlogger.atlassian.net/browse/M2-5267)

This PR adds some new attributes for covering Summary screen with auto-tests

This PR changes:
-  adds some new aqa attributes

### DIFF
--- a/src/entities/alert/ui/Alert.tsx
+++ b/src/entities/alert/ui/Alert.tsx
@@ -9,7 +9,10 @@ function Alert({ children }: Props) {
   return (
     <XStack alignItems="center" mr={25}>
       <Image mt={2} mr={10} width={13} height={16} src={bellAlertIcon} />
-      <Text fontSize={16}>{children}</Text>
+
+      <Text fontSize={16} accessibilityLabel="alert">
+        {children}
+      </Text>
     </XStack>
   );
 }
--- a/src/entities/alert/ui/AlertList.tsx
+++ b/src/entities/alert/ui/AlertList.tsx
@@ -16,7 +16,13 @@ function AlertList({ alerts }: Props) {
   }
 
   return (
-    <Box bg="$lighterGrey4" br={10} p={20} mb={20}>
+    <Box
+      bg="$lighterGrey4"
+      br={10}
+      p={20}
+      mb={20}
+      accessibilityLabel="alerts-group"
+    >
       <Text fontWeight="600" fontSize={23} mb={10}>
         {t('activity_summary:alerts')}
       </Text>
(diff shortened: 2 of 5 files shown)


farmerpaul (commented):
I'm confused why `accessibilityLabel` is being used throughout the mobile app (including in this PR) as the prop to use to target elements in automated testing. I've always used `testID` for this purpose. I understand that the purpose of `accessibilityLabel` is for actual accessibility purposes, i.e., screen readers such as VoiceOver (iOS) and TalkBack (Android). The text in this label is what people with accessibility needs receive in order to navigate the app.

In my experience, `accessibilityLabel` should only be overridden cautiously (usually screen readers read out the text contained in the element, which is a good default), and its value should always be translated using i18next to make sure it's internationalized.

See an example recording of how VoiceOver works in the mobile app currently (make sure to unmute the video below). The screen reader is reporting what should only be used as internal identifiers used strictly by automated tests; these identifiers should not be exposed to people with accessibility needs. In fact, the app is now _less_ accessible because we're overriding the actual text value of calendar weekdays and dates (in this example) with `calendar-week-name` and `calendar-day-of-month`. Normally, VoiceOver would just report the actual label when navigating these elements, rather than an internal identifier.

https://github.com/ChildMindInstitute/mindlogger-app-refactor/assets/1282772/e262af00-7c38-4c98-b2ba-69cfa5d8122c

Perhaps this PR isn't the best place to bring this up, but I'd welcome your thoughts!

anq83:
@farmerpaul  Thanks for your comment. Good point.
Please review the ticket:

Probably, here's the reason:
image

farmerpaul (approved):
Gotcha, thanks for those insights @anq83. I think I understand the issue with Appium; it also seems to be explained pretty well in this [blog post](https://blog.stackademic.com/appium-inspector-demystified-a-guide-to-selecting-elements-on-react-native-2cd3e3c60809). It's unfortunate that `testID` is not handled consistently between iOS and Android.

From a code simplicity perspective, it makes sense to use `accessibilityLabel` to target elements in automated tests. But the downside is that `accessibilityLabel` is also exposed to screen readers, so if we are concerned about making the product accessible, then those labels all need to be (a) internationalized, and (b) should clearly describe the element they are attached to (rather than be written in "developer-speak").

If a11y becomes a product priority down the road, we may have to look into updating these, but for now this is fine. Approving 🙂

BamMironov (approved):

PR #634, opened 2024-02-23
M2-5380: Conditional logic improvement

vmkhitaryanscn:
### 📝 Description

🔗 [Jira Ticket M2-5380](https://mindlogger.atlassian.net/browse/M2-5380)

This PR adds optimizations for conditional logic
Calling `const currentStorageRecord = getCurrentActivityStorageRecord()!` on every iteration on bigger pipelines was causing lags.

This PR changes:
-  Conditional logic improvement

### DIFF
--- a/src/features/pass-survey/model/hooks/useActivityState.ts
+++ b/src/features/pass-survey/model/hooks/useActivityState.ts
@@ -174,7 +174,7 @@ function useActivityState({
   function iterateWithConditionalLogic(fromStep: number) {
     const currentStorageRecord = getCurrentActivityStorageRecord()!;
     const answers = { ...(currentStorageRecord.answers ?? {}) };
-    const timers = { ...(currentStorageRecord.timers ?? {}) };
+    const timers = currentStorageRecord?.timers;
     const items = currentStorageRecord.items ?? {};
 
     for (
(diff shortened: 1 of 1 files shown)


farmerpaul (approved):
I compared performance of progressing through questions of the Healthy Brain Network activities between `dev` and your PR, and it's definitely much faster in your PR – nice work! Code changes look sensible to me so I'll approve, however I think it would be wise for someone more familiar with the code to approve as well.

BamMironov (approved):
Looks much better than the previous implementation. Thanks

src/features/pass-survey/model/hooks/useActivityState.ts
```diff
@@ -171,43 +171,31 @@ function useActivityState({
     });
   }
 
-  function removeAnswer(step: number) {
-    const currentStorageRecord = getCurrentActivityStorageRecord();
-
-    if (!currentStorageRecord) {
-      return;
-    }
+  function removeConditionallyHiddenItemsAnswersAndTimers(fromStep: number) {
```
anq83, line 174: I like the idea to combile callback and iteratePipeline, but 
1. let's keep the previous name - iteratePipeline or iterateWithConditionalLogic
2. it seems this snippet is not considered:
```
if (!currentStorageRecord?.timers) {
   return;
}
```

anq83 (changes requested):

src/features/pass-survey/model/hooks/useActivityState.ts
anq83: 2. yes, but not the same. In the prev. version it was just return, so {} wasn't recorder into the object unlike the current version. 
For example: !!undefined - false, !!{} - true.

anq83 (approved):

PR #636, opened 2024-02-26
M2-4990: Mixpanel Tracking: Notification,reminders, and alerts

anq83:
### 📝 Description

🔗 [Jira Ticket M2-4990](https://mindlogger.atlassian.net/browse/M2-4990)

This PR adds the mix panel event when user taps on event-based notification / reminder

Changes include:

- Extended the enum MixEvents, added new event there, and included into the notification tap handler

### DIFF
--- a/src/shared/lib/analytics/AnalyticsService.ts
+++ b/src/shared/lib/analytics/AnalyticsService.ts
@@ -19,6 +19,7 @@ let service: IAnalyticsService;
 export const MixProperties = {
   AppletId: 'Applet ID',
   MindLoggerVersion: 'MindLogger Version',
+  NotificationType: 'Notification Type',
 };
 
 export const MixEvents = {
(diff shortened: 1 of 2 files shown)


src/shared/lib/analytics/AnalyticsService.ts
```diff
@@ -39,6 +39,7 @@ export const MixEvents = {
   UploadLogsPressed: 'Upload Logs Pressed',
   UploadedLogsSuccessfully: 'Uploaded Logs Successfully',
   UploadLogsError: 'Upload Logs Error Occurred',
+  NotificationTap: 'Notification tap',
```
binho, line 42: Maybe update to `Notification tapped` to match the copy style from the other events?

BamMironov: @binho this is the expected name according to the ticket

BamMironov (approved):
Looks good to me

PR #641, opened 2024-02-27
M2-4794: Added stream ip address and port configuration to applet streaming details

vmkhitaryanscn:
### 📝 Description

🔗 [Jira Ticket M2-4794](https://mindlogger.atlassian.net/browse/M2-4794)

This PR adds stream ip address and port configuration to applet streaming details

This PR changes:
-  Streaming connection modal default values are now filled with default ip address and port , if provided in the admin panel , and if other values are not stored in local store.
- Disconnecting socket connection(if any) when user logs out


BamMironov (approved):
Looks good to me

PR #652, opened 2024-03-04
M2-5595: remove reset password input error message

jgdev:
### 📝 Description

🔗 [Jira Ticket M2-5595](https://mindlogger.atlassian.net/browse/M2-5595)

Changes include:

- Remove input error message from backend and only display the banner error

### DIFF
--- a/src/features/forgot-password/ui/ForgotPasswordForm.tsx
+++ b/src/features/forgot-password/ui/ForgotPasswordForm.tsx
@@ -11,8 +11,8 @@ import {
   useBanner,
   useFormChanges,
 } from '@shared/lib';
-import { YStack, Box, BoxProps, SubmitButton } from '@shared/ui';
-import { ErrorMessage, InputField } from '@shared/ui/form';
+import { Box, BoxProps, SubmitButton } from '@shared/ui';
+import { InputField } from '@shared/ui/form';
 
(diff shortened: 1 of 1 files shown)


src/features/forgot-password/ui/ForgotPasswordForm.tsx
```diff
@@ -60,21 +59,13 @@ const ForgotPasswordForm: FC<Props> = props => {
...
           />
-
-          {error && (
-            <ErrorMessage
-              mode="light"
-              error={{ message: error.evaluatedMessage! }}
-              mt={8}
-            />
-          )}
-        </YStack>
```
ChaconC, line 77: This is simply removing the error message but it's not looking at the source of the problem.

Did you validate if users are able to reset their password?

binho: I think with the fix on BE here https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1146 we can actually keep the error message and close this PR?

mbanting: @jgdev please go ahead and close this PR. This is a [BE issue](https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1146) and has been resolved.

PR #653, opened 2024-03-05
M2-5369, M2-5370 unit tests for alerts-extractor: extract-from-item and penetration tests

anq83:
### 📝 Description

🔗 [Jira Ticket M2-5369](https://mindlogger.atlassian.net/browse/M2-5369)
🔗 [Jira Ticket M2-5370](https://mindlogger.atlassian.net/browse/M2-5370)

Changes include:

- Unit tests for AlertsExtractor -> extractFromItem function
- Unit tests for AlertsExtractor, penetration
- Unit tests minor improvements

### DIFF
--- a/src/features/pass-survey/model/tests/AlertsExtractor.extractFromItem.test.ts
+++ b/src/features/pass-survey/model/tests/AlertsExtractor.extractFromItem.test.ts
@@ -22,15 +22,6 @@ const mockExtractFunctions = (extractor: AlertsExtractor) => {
   extractor.extractFromStackedCheckbox = extractFromStackedCheckboxMock;
   extractor.extractFromStackedRadio = extractFromStackedRadioMock;
   extractor.extractFromStackedSlider = extractFromStackedSliderMock;
-
-  return {
-    extractFromRadioMock,
-    extractFromCheckboxMock,
-    extractFromStackedRadioMock,
-    extractFromStackedCheckboxMock,
-    extractFromSliderMock,
-    extractFromStackedSliderMock,
-  };
 };
 
 describe('AlertsExtractor: test extractFromItem', () => {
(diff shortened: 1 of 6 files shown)


src/features/pass-survey/model/tests/AlertsExtractor.extractFromItem.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  return {
+    extractFromRadioMock,
+    extractFromCheckboxMock,
+    extractFromStackedRadioMock,
+    extractFromStackedCheckboxMock,
+    extractFromSliderMock,
+    extractFromStackedSliderMock,
+  };
+};
```
BamMironov, line 34: I have a couple of questions here:

Why do we need to return the links to the same methods of the extractor object? In my opinion, there is no need to do that at all. We can assign mocking functions to the extractor and then call these functions using their original name (e.g. extractFromCheckbox) in the "it" tests

Can we move the function to the `beforeAll` function and use it instead of `beforeEach`? I don't see a reason why the extractor should be recreated for each tests

Example:
```
...
describe('AlertsExtractor: test extractFromItem', () => {
  let extractor: AlertsExtractor;

  beforeAll(() => {
    extractor = new AlertsExtractor({
      log: jest.fn(),
      warn: jest.fn(),
      error: jest.fn(),
      info: jest.fn(),
    } as unknown as ILogger);

    const extractFromRadioMock = jest.fn();
    const extractFromCheckboxMock = jest.fn();
    const extractFromStackedRadioMock = jest.fn();
    const extractFromStackedCheckboxMock = jest.fn();
    const extractFromSliderMock = jest.fn();
    const extractFromStackedSliderMock = jest.fn();

    extractor.extractFromCheckbox = extractFromCheckboxMock;
    extractor.extractFromRadio = extractFromRadioMock;
    extractor.extractFromSlider = extractFromSliderMock;
    extractor.extractFromStackedCheckbox = extractFromStackedCheckboxMock;
    extractor.extractFromStackedRadio = extractFromStackedRadioMock;
    extractor.extractFromStackedSlider = extractFromStackedSliderMock;
  });

  it('Should call extractFromRadio when type is Radio', () => {
    //@ts-expect-error
    extractor.extractFromItem({ type: 'Radio' } as PipelineItem, {} as Answer);

    expect(extractor.extractFromRadio).toBeCalledTimes(1);
  });
...
```

BamMironov: Maybe, there is no need to even use `beforeAll` at all because the extractor variable is already defined in the scope of the `describe` function

BamMironov (changes requested):
Couple of suggestions provided

src/features/pass-survey/model/tests/AlertsExtractor.extractFromItem.test.ts
anq83: 1. ok, thanks, maube "return" doesn't make sence.
2. Just a good practice for achieving isolation between tests. Theoretically, a tested instance could keep and update any state during tests that'd fail other tests. To analyze and fix such errors would be difficult enough. The goal here is to test an instance's single function in isolation. For the whole component testing, we implement tests that named "penetrating".

PR #656, opened 2024-03-06
M2-4278: [Mobile App] send application logs to server before logging …

anq83:
### 📝 Description

🔗 [Jira Ticket M2-4278](https://mindlogger.atlassian.net/browse/M2-4278)

Changes include:

- Added sending logs before log-out
- Considered repeat if mutex is busy

### DIFF
--- a/src/shared/lib/services/Logger.ts
+++ b/src/shared/lib/services/Logger.ts
@@ -14,6 +14,7 @@ import {
   callWithMutex,
   callWithMutexAsync,
   isAppOnline,
+  wait,
 } from '../utils';
 
 type NamePath = {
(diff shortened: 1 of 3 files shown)


src/shared/lib/services/Logger.ts
```diff
@@ -38,7 +39,7 @@ type DeviceInfoLogObject = {
   lastUpdateTime: number;
 };
 
-class Logger implements ILogger {
+export class Logger implements ILogger {
```
BamMironov, line 42: If this change intends to be able to mock `Logger`'s functions why can't we mock the function of the singletone in the first place?

BamMironov (changes requested):
Looks good but I have one question

src/shared/lib/services/Logger.ts
anq83: To test new instance of the class, not the default one. Again - tests isolation

PR #662, opened 2024-03-12
fix(M2-5809): Fixed stream port pre-filling issue for Live connection form

vmkhitaryanscn:
### 📝 Description

🔗 [Jira Ticket M2-5809](https://mindlogger.atlassian.net/browse/M2-5809)

Changes include:

- Casting number to string appletDetails streamPort.

### ✏️ Notes

We were expecting streamPort field from backend to be a string, however it was a number, so we need to cast it to string in the mapper.

### DIFF
--- a/src/entities/applet/model/mappers.ts
+++ b/src/entities/applet/model/mappers.ts
@@ -89,7 +89,7 @@ export function mapAppletDetailsFromDto(
     encryption: detailsDto.encryption,
     streamEnabled: detailsDto.streamEnabled,
     streamIpAddress: detailsDto.streamIpAddress,
-    streamPort: detailsDto.streamPort,
+    streamPort: detailsDto?.streamPort?.toString() ?? null,
   };
 }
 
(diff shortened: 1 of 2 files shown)


src/entities/applet/model/mappers.ts
```diff
@@ -89,7 +89,7 @@ export function mapAppletDetailsFromDto(
     encryption: detailsDto.encryption,
     streamEnabled: detailsDto.streamEnabled,
     streamIpAddress: detailsDto.streamIpAddress,
-    streamPort: detailsDto.streamPort,
+    streamPort: detailsDto.streamPort ? `${detailsDto.streamPort}` : null,
```
BamMironov, line 92: If the intent is to cast a number value to string then please consider this piece of code code.
```suggestion
    streamPort: detailsDto?.streamPort?.toString() ??  null,
```

BamMironov (changes requested):

BamMironov (approved):

PR #665, opened 2024-03-13
M2:5065: Implement password reset form and deep linking

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-5065](https://mindlogger.atlassian.net/browse/M2-5065)

This PR is a combination of two changes: A password recovery form, and the introduction of deep linking.

#### Password recovery

This is just your standard password recovery implementation, very similar to https://github.com/ChildMindInstitute/mindlogger-admin/pull/1510. Some notable things about the implementation:

- I bumped the version of [react-native-progress](https://github.com/oblador/react-native-progress) to 5.0.1 because it has a [bug fix](https://github.com/oblador/react-native-progress/releases/tag/v5.0.1) for the circle progress indicator, which I've created a reusable component for and make use of.
- I added a banner to the login screen upon successful password recovery

#### Deep Linking

I've configured the mobile apps to respond to links on the respondent web app domain. Initially this only works for the `/password-recovery` path to satisfy this story, but the foundation has been laid to intercept other paths in the future. The app will fallback to the device's default browser if there are no matching screens for a given path.

The respondent web app domain has to be provided at build time, so you'll need to update your `.env.dev` file with the following value

https://github.com/ChildMindInstitute/mindlogger-app-refactor/blob/246f631b00d30e89d634ae718e2989968f17aaa6/.env.example#L5

In order to successfully implement this feature, a number of related things need to be done

#### App Links

I opted to do this in the app links format, which means the mobile apps will intercept a URL with an `https` scheme. In addition to the mobile apps specifying which domain they'd like to intercept, both Apple and Google require that the domain proves association with the mobile apps by serving a static JSON file from a well known path. On iOS, this is at `/.well-known/apple-app-site-association`, and on Android this is at `/.well-known/assetlinks.json`. I've also created a PR at the respondent web app repository (https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/406) that needs to be merged in order for this PR to be tested successfully.

In addition to the PR I mentioned above, the deployment process for the respondent web app will need to be updated to provide the correct values in the JSON files for each environment so that it will be linked to the correct app.

> [!NOTE]
> This has since been completed ✅

#### Apple Developer Capabilities

In order for the site association file to be respected by iOS, the "Associated Domains" capability needs to be enabled for all the various builds of the app. It has thus far been enabled for the dev build, but it needs to be enabled for the other builds as well.

> [!NOTE]
> Please verify that the domains I've used for each environment are correct!

### 📸 Screenshots
Screenshot 2024-03-13 at 12 52 45 PM Screenshot 2024-03-13 at 12 52 45 PM Screenshot 2024-03-13 at 12 52 45 PM

### 🪤 Peer Testing

These steps assume you have the backend API running. Check out the branch locally, then be sure to add the new environment variable

https://github.com/ChildMindInstitute/mindlogger-app-refactor/blob/246f631b00d30e89d634ae718e2989968f17aaa6/.env.example#L5

Clean your iOS build folder

```shell
yarn clean-ios
```

Start metro and launch the iOS simulator

```shell
# Long running process
yarn start:reset-cache

# In a separate terminal tab
yarn ios
```

Once the app has launched, you'll want to test a few scenarios. There's no direct link to the password reset page, so you get to test it in tandem with the deep linking.

#### Invalid domain

In this scenario, you try to launch an unrelated domain. You should observe that it launches in Safari

```shell
# You may be prompted to install uri-scheme
npx uri-scheme open "https://example.com" --ios
```

image

#### Invalid path

In this scenario, you try to launch the respondent web app at an unsupported path. You should observe that it launches in Safari

```shell
npx uri-scheme open "https://web-dev.cmiml.net/login" --ios
```

image

#### Invalid query params

In this scenario, you try to launch the respondent web app at the `/password-recovery` path but with an invalid `email` and/or `key` query param. You should observe that the app is launch to the invalid link screen

```shell
npx uri-scheme open "https://web-dev.cmiml.net/password-recovery?email=someone&key=abc" --ios
```

https://github.com/ChildMindInstitute/mindlogger-app-refactor/assets/14842108/302b41ce-2953-443a-aeed-eba1a5098b61

#### Valid password reset link

In this scenario, you will go to the forgot password screen (accessible from the login screen) and request a password reset link. Access the email in MailHog (localhost:8025), copy the link, and edit the domain to match `web-dev.cmiml.net`. Then execute the following command:

```shell
npx uri-scheme open "https://web-dev.cmiml.net/password-recovery?key=16f021ea-9047-301b-be72-49db114f5398&email=kenroy.gobourne%40metalab.com" --ios
```

You should observe that it goes to the password recovery form.

https://github.com/ChildMindInstitute/mindlogger-app-refactor/assets/14842108/d8e48a50-437a-4a19-abb6-d6873c603e77

At this point you may repeat this step to test the functionality of the form


assets/translations/fr.json
```diff
@@ -142,6 +142,20 @@
     "change_pass": "Changer le mot de passe",
     "password_at_least_characters": "Le mot de passe doit être au moins de 6 caractères"
   },
+  "password_recovery_form": {
+    "password_at_least_characters": "Le mot de passe doit être au moins de 6 caractères",
```
binho, line 146: If it's easy: can this `6` be replaced with the constant we use when checking the password?

binho (approved):
LGTM! Just added a small comment with a possible future-prof improvement.

src/features/password-recovery/model/PasswordRecoveryFormSchema.ts
```diff
@@ -0,0 +1,22 @@
...
+  .object({
+    new_password: z
+      .string()
+      .trim()
+      .min(6, 'password_recovery_form:password_at_least_characters')
+      .refine(
+        value => !value.includes(' '),
+        'password_recovery_form:password_no_spaces',
+      ),
+    confirm_password: z
```
BamMironov, line 13: Please use camel cause for field names

src/features/password-recovery/ui/PasswordRecoveryForm.tsx
```diff
@@ -0,0 +1,116 @@
...
+        duration: 5000,
+      });
+    },
+  });
+
+  const { form, submit } = useAppForm(PasswordRecoveryFormSchema, {
+    defaultValues: {
+      new_password: '',
+      confirm_password: '',
+    },
```
BamMironov, line 54: Please use camel cause for field names

src/app/ui/AppProvider/NavigationProvider.tsx
```diff
@@ -15,10 +21,33 @@ const theme = {
   },
 };
 
+const linking: LinkingOptions<ReactNavigation.RootParamList> = {
+  prefixes: [DEEP_LINK_PREFIX!],
+  getStateFromPath: (path, options) => {
+    const state = getStateFromPath(path, options);
```
BamMironov, line 27: Please make sure that parameters (e.g. email) are parsed correctly. Sometimes they don't if a parameter contains special characters. Wrap in `decodeURIComponent` if needed

BamMironov (changes requested):
I left some comments from my side.
Overall, nice job!

src/app/ui/AppProvider/NavigationProvider.tsx
```diff
@@ -15,10 +21,33 @@ const theme = {
   },
 };
 
+const linking: LinkingOptions<ReactNavigation.RootParamList> = {
+  prefixes: [DEEP_LINK_PREFIX!],
```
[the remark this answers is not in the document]
sultanofcardio: I already check if it is `undefined` here:

https://github.com/ChildMindInstitute/mindlogger-app-refactor/blob/a7dac0241ab5d1862ad029b8b3faebd627151709/src/app/ui/AppProvider/NavigationProvider.tsx#L50

So the linking config is only provided if it is not `undefined`. I can update this to make an error log, but I'm not sure what the benefit would be. Do we receive error logs from the mobile app in something like CloudWatch?

src/app/ui/AppProvider/NavigationProvider.tsx
sultanofcardio: Since this is coming from a generated email, and the BE already encodes the query params I made the assumption this would work fine as this is how it's done on the other FE apps. Do you have an example that fails the parsing?

src/app/ui/AppProvider/NavigationProvider.tsx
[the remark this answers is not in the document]
BamMironov: @sultanofcardio We do collect logs. Right now, we don't use CloudWatch but S3 to store them.
Logs have helped us a lot to investigate bugs that are hard to reproduce or in case we have little information from users. So I think it would be beneficial to add them for deep links too.

src/app/ui/AppProvider/NavigationProvider.tsx
BamMironov: To be honest, I don't have any specific examples of something that can go wrong from the UI side. However, I've got plenty of examples when the backend broke the parsing on their side which resulted in breaking of deep linking. Better safe than sorry.

src/app/ui/AppProvider/NavigationProvider.tsx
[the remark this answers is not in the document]
sultanofcardio: Makes sense to me. Gonna add an error log in there 👍

src/app/ui/AppProvider/NavigationProvider.tsx
sultanofcardio: Sounds good. I'll add it in

sultanofcardio: Ok I've done some testing around this and I can't get this to crash with any special characters. However, `decodeURIComponent` does crash if the argument isn't properly URL encoded. I'm willing to merge this in as-is, as trying to handle special characters is leading to more errors. What do you think?

PR #674, opened 2024-03-15
M2-5569, M2-5879: Keep data upon logout & loading view

BamMironov:
### 📝 Description

🔗 [Jira Ticket M2-5569](https://mindlogger.atlassian.net/browse/M2-5569)
🔗 [Jira Ticket M2-5569](https://mindlogger.atlassian.net/browse/M2-5879)

Changes include:

- Rebuilt logout process so that it doesn't clear MMKV storages and redux state
- Rebuilt login process so that it checks if another user logs in. If so, the system clears MMKV storages.
- Loading view on the Settings screen that shows basic UI (spinner) and blocks user interaction during the logout process.

### 📸 Screenshots
Android
telegram-cloud-photo-size-2-5451778082928186603-y

iOS
image


anq83 (approved):
Please, ensure that all the data that cleared in the current logout will be restored during login even if user is not changed.
I mean react-query cache, images cache, session-model, tcp-connection

PR #675, opened 2024-03-18
(temp brach) M2-5734 autocompletion mvp

anq83:
M2-5734 autocompletion mvp (temp brach)


src/screens/model/checkEntityAvailability.ts
```diff
@@ -34,10 +34,23 @@ export const checkEntityAvailability = ({
...
 
+  const progressPayload = storeProgress[appletId]?.[entityId]?.[eventId];
+
+  const isInProgress =
+    !!progressPayload && progressPayload.startAt && !progressPayload.endAt;
+
+  const shouldBeAutocompleted =
+    isInProgress &&
+    progressPayload.availableTo &&
+    progressPayload.availableTo < Date.now();
```
anq83, line 45: Check if seconds equal to 59

src/widgets/survey/model/hooks/useFlowState.ts
```diff
@@ -25,10 +27,22 @@ export function useFlowState({ appletId, eventId, flowId }: UseFlowStateArgs) {
   const flowSummaryData: FlowSummaryData = (record?.context?.[SummaryDataKey] ??
     {}) as FlowSummaryData;
 
+  const restFlowActivityIds: string[] = useMemo(() => {
```
BamMironov, line 30: Please rename to `remainingActivityIds`. The "flow" word can be omitted due to the fact that this is already in the scope of use**Flow**State function

src/widgets/survey/model/hooks/useAutoCompletion.ts
```diff
@@ -0,0 +1,205 @@
...
+type Result = {
+  process: ProcessAutocompletion;
+  completeEntityIntoUploadToQueue: CompleteEntityIntoUploadToQueue;
+  hasItemsInQueue: boolean;
+};
+
+export const AutoCompletionMutex = Mutex();
+
+const useAutoCompletion = (): Result => {
+  const mutex = useRef(AutoCompletionMutex).current;
```
BamMironov, line 41: Please use `AutoCompletionMutex` directly instead. There is no need to use it through ref.current.

`const mutex = useRef(AutoCompletionMutex).current;`
is the same as
`const mutex = AutoCompletionMutex;`

src/widgets/survey/model/hooks/useAutoCompletion.ts
```diff
@@ -0,0 +1,205 @@
...
+    try {
+      if (collectOutput.completionType === 'intermediate') {
+        await constructService.constructForIntermediate({
+          ...collectOutput,
+          flowId: collectOutput.flowId!,
+        });
+      }
+
+      if (collectOutput.completionType === 'finish') {
+        await constructService.constructForFinish(collectOutput);
```
BamMironov, line 82: Since `completionType` is a field of `collectOutput` and `collectOutput` is passed down to `constructService`, can `constructService` make the decision of what constructor (constructForIntermediate, constructForFinish) function to call internally?

BamMironov (changes requested):

BamMironov:
The changes were moved to https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/697

PR #697, opened 2024-03-29
M2-5734: [Mobile] - In Progress Activities and Flows Auto-Completion

anq83:
### 📝 Description

🔗 [Jira Ticket M2-5734](https://mindlogger.atlassian.net/browse/M2-5734)

Changes include:

- Autocompletion services and their integration
- Autocompletion triggers by the same triggers as the ones that made upload queue
- Also, the new services integrated into useStartEntity hook, so that the (re)starting/resuming entity will be aucompleted if it's expiried


src/entities/notification/model/hooks/useOnNotificationRefresh.ts
```diff
@@ -0,0 +1,54 @@
...
+      NotificationModel.NotificationRefreshService.refresh(
+        queryClient,
+        storeProgress,
+        completions,
+        logTrigger!,
+      )
+        .finally(() => {
+          Logger.send();
+        })
+        .finally(() => {
```
BamMironov, line 36: Please join the `finally` blocks into one

BamMironov (changes requested):
One comment from my side.
In general, amazing work!

src/entities/notification/model/hooks/useOnNotificationRefresh.ts
anq83: ok

BamMironov (approved):

PR #710, opened 2024-04-03
M2-5017: A/B Trails - Save Incomplete Round of the Test and Continue

vmkhitaryanscn:
### 📝 Description

🔗 [Jira Ticket M2-5017](https://mindlogger.atlassian.net/browse/M2-5017)

Changes include:

- A/B Trails - Save Incomplete Round of the Test and Continue feature implementation

### ✏️ Notes

Added 2 new types of userActions, added maximumIndex to AbTrails response.

### DIFF
--- a/src/entities/abTrail/ui/AbCanvas.tsx
+++ b/src/entities/abTrail/ui/AbCanvas.tsx
@@ -285,6 +285,12 @@ const AbCanvas: FC<Props> = props => {
         x.actual = mark.actual ?? null;
       }
     });
+
+    onLogResult({
+      lines: logLines,
+      currentIndex: getCurrentIndex() + 1,
+      maximumIndex: testData.nodes.length,
+    });
   };
 
   const getGreenPointIndex = () => {
@@ -354,10 +360,6 @@ const AbCanvas: FC<Props> = props => {
       addOverCorrectPointToStream(point);
       keepPathInState();
       resetCurrentPath();
-      onLogResult({
-        lines: logLines,
-        currentIndex: getCurrentIndex() + 1,
-      });
       onMessage(MessageType.Completed);
       onComplete();
 
--- a/src/entities/abTrail/ui/AbCanvas.tsx
+++ b/src/entities/abTrail/ui/AbCanvas.tsx
@@ -285,7 +285,9 @@ const AbCanvas: FC<Props> = props => {
         x.actual = mark.actual ?? null;
       }
     });
+  };
 
+  const onResult = () => {
     onLogResult({
       lines: logLines,
       currentIndex: getCurrentIndex() + 1,
@@ -357,6 +359,7 @@ const AbCanvas: FC<Props> = props => {
 
     if (isOverNext(point) && isOverLast(point)) {
       markLastLogPoints({ valid: true });
+      onResult();
       addOverCorrectPointToStream(point);
       keepPathInState();
       resetCurrentPath();
@@ -368,6 +371,7 @@ const AbCanvas: FC<Props> = props => {
 
     if (isOverNext(point)) {
       markLastLogPoints({ valid: true });
+      onResult();
       addOverCorrectPointToStream(point);
       keepPathInState();
       reCreatePath(point);
@@ -379,6 +383,7 @@ const AbCanvas: FC<Props> = props => {
     if (isOverWrong(point)) {
       const node = findNodeByPoint(point)!;
       markLastLogPoints({ valid: false, actual: node.label });
+      onResult();
       setErrorPath(currentPath);
       resetCurrentPath();
       setFlareGreenPointIndex({ index: getCurrentIndex() });
@@ -409,6 +414,7 @@ const AbCanvas: FC<Props> = props => {
     addOverUndefinedPointToStream(point);
 
     markLastLogPoints({ valid: false, actual: node?.label ?? 'none' });
+    onResult();
   };
 
   const drawPath = () => {
--- a/src/entities/abTrail/ui/AbCanvas.tsx
+++ b/src/entities/abTrail/ui/AbCanvas.tsx
@@ -356,14 +356,13 @@ const AbCanvas: FC<Props> = props => {
       resetCloseToNextRerendered();
     }
 
-    onResult();
-
     if (isOverNext(point) && isOverLast(point)) {
       markLastLogPoints({ valid: true });
       addOverCorrectPointToStream(point);
       keepPathInState();
       resetCurrentPath();
       onMessage(MessageType.Completed);
+      onResult();
       onComplete();
 
       return;
@@ -388,6 +387,7 @@ const AbCanvas: FC<Props> = props => {
       onMessage(MessageType.IncorrectLine);
 
       addOverWrongPointToStream(point, node.label);
+      onResult();
       return;
     }
 
--- a/src/entities/abTrail/ui/AbCanvas.tsx
+++ b/src/entities/abTrail/ui/AbCanvas.tsx
@@ -36,7 +36,7 @@ type Props = {
   testData: AbTestPayload;
   readonly: boolean;
   width: number;
-  onLogResult: (data: OnResultLog) => void;
+  onResult: (data: OnResultLog) => void;
   onMessage: (message: MessageType) => void;
   onComplete: () => void;
 } & StreamEventLoggable<AbTestStreamEvent> &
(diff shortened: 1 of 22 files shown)


BamMironov (changes requested):
A couple of comments from my side regarding checking if the code needs to show the new popup

src/features/pass-survey/lib/services/SkipService.ts
BamMironov, line 1: Please create two public methods: `skip` and `proceed`. They will call either `onSkip` or `onProceed`. But it may contain additional logic in the future, which is why it's better to have wrapper functions.

Also, create a private method `canSkipABTest` that will be called if `itemType` is `AbTest` and move the relevant logic from `canSkip` method.
Maybe would be better to add a switch case to `canSkip`

src/features/pass-survey/ui/ActivityStepper.tsx
```diff
@@ -177,6 +187,24 @@ function ActivityStepper({
...
+      )
+    ) {
+      trackUserAction(userActionCreator.next());
+      const shouldSkipRound = await fetchSkipActivityUserConfirmation();
+
+      if (shouldSkipRound) {
+        skipService.onSkip();
+        return 1;
+      } else {
+        skipService.onProceed();
```
BamMironov, line 203: and here you could call skipService.skip() and skipService.proceed();

src/features/pass-survey/lib/services/SkipService.ts
```diff
@@ -0,0 +1,30 @@
+import { AbTestResponse, Answer, PipelineItem } from '@features/pass-survey';
+
+type Params = {
+  onSkip: () => void;
+  onProceed: () => void;
+};
+
+class SkipService {
+  public onSkip;
+  public onProceed;
```
BamMironov, line 10: Please make them private and add types.

src/entities/abTrail/ui/AbCanvas.tsx
```diff
@@ -351,13 +359,10 @@ const AbCanvas: FC<Props> = props => {
```
BamMironov, line 359: I wonder if it's possible to just add onResult() here, before all the if-statements

src/entities/abTrail/ui/AbCanvas.tsx
```diff
@@ -287,6 +287,14 @@ const AbCanvas: FC<Props> = props => {
     });
   };
 
+  const onResult = () => {
```
BamMironov, line 290: I think, there has been an issue with the naming of AbCanvas' `onLogResult` function. It should have been named `onResult` in the first place.
Regarding this, onResult function, I think it could stay as-is. But in order to avoid naming confusion I suggest calling  `props.onResult(...)` inside this function.

BamMironov (changes requested):

BamMironov (approved):

src/shared/ui/Stepper/NextButton.tsx
```diff
@@ -12,18 +14,29 @@ type Props = PropsWithChildren<{
...
+const THROTTLE_DELAY = 300;
+
 function NextButton({ children, isIcon, accessibilityLabel }: Props) {
   const { next } = useContext(HandlersContext);
 
   const onPressNext = useCallback(() => {
     next({ isForced: false, shouldAutoSubmit: true });
   }, [next]);
 
+  const onPressNextDebounced = useThrottledCallback(
```
anq83, line 26: Can you explain necessarity of this, please?

anq83 (changes requested):
Review in progress, I'll ask different things via comments from time to time.

src/features/pass-survey/model/hooks/useActivityStepper.ts
```diff
@@ -15,10 +15,18 @@ function useActivityStepper(state: ActivityState | undefined) {
   const step = state?.step ?? 0;
   const items = state?.items ?? [];
   const answers = state?.answers ?? {};
+  const actions = state?.actions ?? [];
 
+  const lastAction = actions[actions.length - 1];
   const currentPipelineItem = state?.items[step];
+  const nextPipelineItem = state?.items[step + 1];
+
+  const shouldSkipNextUserAction =
```
anq83, line 24: Please rename to "shouldIgnoreUserActionTrack" as we've already Skip action and this variable name confusing

src/features/pass-survey/ui/ActivityStepper.tsx
```diff
@@ -81,6 +90,7 @@ function ActivityStepper({
     isFirstStep,
     isTutorialStep,
     isConditionalLogicItem,
+    shouldSkipNextUserAction,
```
anq83, line 93: Please rename to "shouldIgnoreUserActionTrack"

src/features/pass-survey/ui/ActivityStepper.tsx
```diff
@@ -136,7 +146,7 @@ function ActivityStepper({
 
     if (canSkip) {
       trackUserAction(userActionCreator.skip());
-    } else {
+    } else if (!shouldSkipNextUserAction) {
```
anq83, line 149: Indeed it looks like workaround. We build BL based on user actions! 
Can we change the solution to another one.
For example: pass ignoreUserActionTrack as the parameter of onNext, and evaluate this value in the onBeforeNext?

anq83 (changes requested):

src/features/pass-survey/lib/services/SkipService.ts
```diff
@@ -1,13 +1,17 @@
 import { AbTestResponse, Answer, PipelineItem } from '@features/pass-survey';
 
+type NoParamVoidFunction = () => void;
```
anq83, line 3: Please move to abstract folder

src/features/pass-survey/lib/services/SkipService.ts
```diff
@@ -1,13 +1,17 @@
...
+
+type PipelineItemType = PipelineItem['type'];
```
anq83, line 5: Please move it to where PipelineItem is placed

src/features/pass-survey/ui/ActivityStepper.tsx
```diff
@@ -158,9 +161,12 @@ function ActivityStepper({
     trackUserAction(userActionCreator.back());
   };
 
-  const onBeforeNext = async (): Promise<number> => {
+  const onBeforeNext = async (): Promise<{
+    stepShift: number;
+    shouldIgnoreUserActionTrack?: boolean;
```
anq83, line 166: We don't need and don't pass shouldIgnoreUserActionTrack into the onBeforeNext

anq83 (changes requested):

src/features/pass-survey/lib/services/SkipService.ts
BamMironov: @anq83 
Sorry, are we even bothering to create a separate type for this?
`() => void` is completely understandable and even shorter than `NoParamVoidFunction`

src/shared/ui/Stepper/index.tsx
```diff
@@ -80,14 +86,15 @@ export function Stepper({
...
+      const { stepShift = 1, shouldIgnoreUserActionTrack = false } =
+        (await onBeforeNextRef.current?.(step)) ?? {};
       const nextStep = step + stepShift;
 
       const moved = viewSliderRef.current?.next(stepShift);
 
       if (moved) {
         stepRef.current = nextStep;
-        onNextRef.current?.(nextStep, isForced);
+        onNextRef.current?.(nextStep, isForced, shouldIgnoreUserActionTrack);
```
BamMironov, line 97: @anq83 I am sorry but I don't think it was a good idea to let the shared component know about "actions". The only things that Stepper should know and operate are steps and isForced.

@vmkhitaryanscn 
Please make it possible to pass a TPayload object to the function:

```
...
  onNext?: (
    step: number,
    isForced: boolean,
    payload: TPayload,
  ) => void;
  onBeforeNext?: (
    step: number,
  ) => Promise<{ stepShift: number; payload: TPayload }>;
...
const { stepShift = 1, payload } = (await onBeforeNextRef.current?.(step)) ?? {};
...
onNextRef.current?.(nextStep, isForced, payload);
```
 
Then pass the payload in the outer scope (in the ActivityStepper component)

BamMironov (changes requested):

src/shared/ui/Stepper/index.tsx
```diff
@@ -15,21 +15,24 @@ export * from './contexts';
 
 export { default as useOnUndo } from './useOnUndo';
 
+export type TPayload = {
```
BamMironov, line 18: I thought you would use it as a generic. Sorry.
Okay but in this case please rename the type to `StepperPayload`

BamMironov (approved):

anq83 (approved):

PR #716, opened 2024-04-08
M2-5888: [Mobile app] Incorrect data in the activity journey file

vmkhitaryanscn:
### 📝 Description

🔗 [Jira Ticket M2-5888](https://mindlogger.atlassian.net/browse/M2-5888)
🔗 [Jira Ticket M2-6116](https://mindlogger.atlassian.net/browse/M2-6116)

Changes include:

- Updated updateUserActionsWithAnswer and updateUserActionsWithAdditionalAnswer to update only the last userAction , as these functions were updating all the occurrences of "SET_ANSWER" events for the item. For now changes only affect TextInput item type.
- [Translation updates](https://mindlogger.atlassian.net/browse/M2-6116)

### DIFF
--- a/assets/translations/en.json
+++ b/assets/translations/en.json
@@ -203,7 +203,7 @@
   "logout_warning": {
     "warning": "Logout Warning",
     "uploads_in_progress": "Uploads in progress",
-    "currently_uploading": "You have responses that are currently uploading. If you log out now your responses will be lost.",
+    "currently_uploading": "You have responses that are currently uploading. If you log out now and then log in as a different user, your responses will be lost.",
     "not_uploaded": "You have not uploaded responses. If you log out now these responses will be lost.",
     "sure_logout": "Are you sure that you want to log out?",
     "cancel": "Cancel",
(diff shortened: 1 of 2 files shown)


src/features/pass-survey/model/hooks/useUserActionManager.ts
```diff
@@ -41,12 +50,14 @@ function useUserActionManager({ activityId, activityState }: Args) {
       | SetAnswerAction
       | undefined;
 
-    const shouldUpdateLastAction =
+    const shouldUpdateAllPreviousAnswerActions =
```
BamMironov, line 53: We need to make sure that we do want to update all the previous `SET_ANSWER` actions in some cases. At the moment, I can't think of an activity item that should correspond to such a rule. Please investigate it and correct me if I am wrong.
If am not wrong then we should always update the last SET_ANSWER action.

BamMironov (changes requested):

BamMironov (approved):

PR #717, opened 2024-04-11
M2-6124, M2-6125, M2-6126: Unit tests for: AvailableToEvaluator, CollectCompletionsService

anq83:
### 📝 Description

Added unit tests that cover AvailableToEvaluator, CollectCompletionsService

🔗 [Jira Ticket M2-6124](https://mindlogger.atlassian.net/browse/M2-6124)
🔗 [Jira Ticket M2-6125](https://mindlogger.atlassian.net/browse/M2-6125)
🔗 [Jira Ticket M2-6126](https://mindlogger.atlassian.net/browse/M2-6126)

### DIFF
--- a/src/widgets/survey/model/services/tests/CollectCompletionsService.collectAll.test.ts
+++ b/src/widgets/survey/model/services/tests/CollectCompletionsService.collectAll.test.ts
@@ -1,0 +1,294 @@
+import { EntityPath } from '@app/abstract/lib';
+import * as survey from '@app/shared/lib/utils/survey/survey';
+
+import { getFlowProgressRecord, getRegularProgressRecord } from './testHelpers';
+import * as storageHelpers from '../../../lib/storageHelpers';
+import { CollectCompletionsService } from '../CollectCompletionsService';
+
+jest.mock('@app/shared/lib/services/Logger', () => ({
+  log: jest.fn(),
+  info: jest.fn(),
+}));
+
+describe('Test CollectCompletionsService: collectAll', () => {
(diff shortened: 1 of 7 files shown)


src/widgets/survey/model/services/tests/CollectCompletionsService.collectAll.test.ts
```diff
@@ -0,0 +1,294 @@
...
+    firstProgress.payload.availableTo = new Date(2023, 3, 7).getTime();
+    secondProgress.payload.availableTo = new Date(2023, 3, 7).getTime();
+
+    const service = new CollectCompletionsService([
+      firstProgress,
+      secondProgress,
+    ]);
+
+    //@ts-expect-error
+    service.collect = jest
```
BamMironov, line 157: Is there a reason to mock the method? I can see it as an important part of the `CollectCompletionsService`, not just a utility method like, for instance, `getNow`

BamMironov (changes requested):

src/widgets/survey/model/services/tests/CollectCompletionsService.collectAll.test.ts
anq83: Yes, it is. The "collect" method and the methods-chain initiated by collect - has been already covered by the tests of another public function: collectForEntity. So, no need to do double work and to make tests more complex. Coverage report for this module shows 100%.

BamMironov (approved):
Looks good to me

PR #718, opened 2024-04-12
M2-6242: Unit tests for activity entity mapToActivity function

vmkhitaryanscn:
### 📝 Description

Added unit tests for activity entity mapToActivity function

🔗 [Jira Ticket M2-6242](https://mindlogger.atlassian.net/browse/M2-6242)


anq83 (changes requested):
1 activity item - 1 input/output mock - 1 test (you may use forEach and the same it block); 
Place mocks into ts file(s) ; 
Review existing tests and apply  the same style ; 
Review coverage of nested functions, % branch

anq83 (changes requested):
Please, add type for every object in the mock file and remove all "as".
const mockObject: TypeOfObject = { ...

PR #721, opened 2024-04-16
M2-6256: [Mobile] Modal "restart-resume" should not appear for in-pro…

anq83:
### 📝 Description

🔗 [Jira Ticket M2-6256](https://mindlogger.atlassian.net/browse/M2-6256)

Changes include:

- Fix the bug - Timer elapse is not managed when tap on notification
- Hide callstack of deeplink settings absence error under DEV mode

### DIFF
--- a/src/app/ui/AppProvider/NavigationProvider.tsx
+++ b/src/app/ui/AppProvider/NavigationProvider.tsx
@@ -26,15 +26,10 @@ const getLinking = ():
   | LinkingOptions<ReactNavigation.RootParamList>
   | undefined => {
   if (!DEEP_LINK_PREFIX) {
-    if (__DEV__) {
-      Logger.info(
-        `[${LOGGER_MODULE_NAME}] No deep link prefix found, deep linking will not work.`,
-      );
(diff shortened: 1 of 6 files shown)


src/app/ui/AppProvider/NavigationProvider.tsx
```diff
@@ -26,9 +26,15 @@ const getLinking = ():
   | LinkingOptions<ReactNavigation.RootParamList>
   | undefined => {
   if (!DEEP_LINK_PREFIX) {
-    Logger.error(
-      `[${LOGGER_MODULE_NAME}] No deep link prefix found, deep linking will not work.`,
-    );
+    if (__DEV__) {
```
BamMironov, line 29: Please avoid chaining if-statements. Therefore, please move this condition to the upper-level condition (`if (!DEEP_LINK_PREFIX)`)

BamMironov (changes requested):
Just one comment from my side. Everything else looks good to me.

BamMironov (approved):

PR #733, opened 2024-04-17
M2-6218: wrong cache directory after app updates

BamMironov:
### 📝 Description

🔗 [Jira Ticket M2-6218](https://mindlogger.atlassian.net/browse/M2-6218)

Changes include:

- The way how files are fetched from the file system. Previously, the app code relied on the pre-saved URI that contained the whole path to the file system. However, the application directory gets renamed upon app update (the content of the directory remains the same). This caused an issue when a file was created before an app updated and then accessed after the app update: the system could not find the file because the file path contained the wrong name of the app directory. This is why it is required to work with file names only and then build the file paths at the exact moment of fetching the files from the file system.
- Better FileSystem mock developed by the author of react-native-file-access
- Unit tests for file.ts utils.


__mocks__/react-native-file-access.ts
BamMironov, line 1: Copy and pasted from https://github.com/alpha0010/react-native-file-access/blob/master/jest/react-native-file-access.ts

src/entities/activity/lib/services/AnswersUploadService.ts
```diff
@@ -116,7 +116,13 @@ class AnswersUploadService implements IAnswersUploadService {
     logAnswerIndex: number,
     appletId: string,
   ): Promise<string> {
-    const localFileExists = await FileSystem.exists(mediaFile.uri);
+    /** Since the application directory changes upon app update
+     * it is required to build URIs from scratch using file names
+     * just before we actually want to fetch those files.
+     * Please see: https://github.com/joltup/rn-fetch-blob/issues/204#issuecomment-786321861
+     */
+    const localFileUri = getLocalFileUri(mediaFile.fileName);
```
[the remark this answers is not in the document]
anq83: Let's remove url property from MediaFile and check the code thoroughly by tsc.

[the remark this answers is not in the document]
anq83: Please, ensure the change is valid for the rest of items: proto, audio / video record.

anq83 (approved):
Please, check the comments and create the separated ticket for refactoring. Refactor is necessary asap.

src/entities/activity/lib/services/AnswersUploadService.ts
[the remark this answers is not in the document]
BamMironov: @anq83
I agree with you on the renaming.
Regarding moving the comment, in my opinion, it is better to keep the comment in the same place as it is now because here we describe the **intent** to use the util function in the scope of the `processFileUpload` method that implements the business logic.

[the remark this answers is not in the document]
anq83: The service where you placed it is complex enough. Let's do not complicate it event more by adding the comment 4-5 rows. Moreover - comments tend to become outdated and the service tends to change from time to time.
To analyze why the function has been applied - developer just clicks on F12.

src/app/model/migrations/migrations/to0002/MigrationUtils0002.ts
```diff
@@ -0,0 +1,78 @@
...
+};
+
+export const evaluateLocalFileUri = (fileName: string) =>
+  `file://${Dirs.CacheDir}/${fileName}`;
+
+export const moveMediaFileToCache = async (
+  fileName: string,
+  fromUri: string,
+) => {
+  const localFileUri = evaluateLocalFileUri(fileName!);
```
anq83, line 40: Please, rename to toUrl

src/app/model/migrations/migrations/to0002/MigrationUtils0002.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  let fileExists = await FileSystem.exists(localFileUri);
+
+  if (fileExists) {
```
anq83, line 44: How is it possible that file exist in the re-named, current cache directory during migration 2 run?
Should migration be runned in this case?
Should we delete files in such cases?
Shoudn't we make check if file exist for fromUrl at the start of this function?

src/app/model/migrations/migrations/to0002/MigrationToVersion0002.ts
```diff
@@ -0,0 +1,164 @@
...
+      if (shouldRenameMediaFile) {
+        renameMediaFile(action.payload.answer.value.answer!);
+      }
+    }
+  }
+
+  private async getUpdatedActivityState(
+    activityFrom: ActivityState,
+  ): Promise<ActivityState> {
+    await Promise.all([
```
anq83, line 121: Maybe it worth to cover every internal function with try-carch to allow to Promise.all to complete their jobs?

And it's defenitelly worth to add more logs in every function about:  pathFrom, pathTo, renameFileNameFrom, renameFileNameTo, "move (un)succesfull" "rename un(succesfull)", an so on, in order to get possibility to analyze logs in case of issues.

anq83 (changes requested):

src/entities/activity/lib/services/AnswersUploadService.ts
```diff
@@ -84,23 +85,32 @@ class AnswersUploadService implements IAnswersUploadService {
...
+  }
+
   private collectFileIds(answers: AnswerDto[]): string[] {
     const result: string[] = [];
 
     for (const itemAnswer of answers) {
       const answerValue = (itemAnswer as ObjectAnswerDto)?.value;
 
-      const mediaAnswer = answerValue as MediaFile;
+      const mediaAnswer = answerValue as FileDto;
```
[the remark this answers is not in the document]
BamMironov: Done

src/shared/lib/utils/file.ts
```diff
@@ -9,3 +11,13 @@ export const isLocalFileUrl = (value: string) => {
 
 export const getFilePath = (path: string) =>
   IS_ANDROID ? `file://${path}` : path;
+
+export const evaluateLocalFileUri = (fileName: string) =>
+  `file://${Dirs.CacheDir}/${fileName}`;
+
+export const getFilenameFromLocalUri = (localUri: string) => {
```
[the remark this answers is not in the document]
BamMironov: already there

src/app/model/migrations/migrations/to0002/MigrationUtils0002.ts
BamMironov: What is migration 2 run?

BamMironov: The function was copy-pasted from the mediaFiles.ts for safety reasons as we would not want to make different implementations of the same function in migration scripts.
Answering your question, it is not possible that the file already exists in cacheFileUri during the migration.

src/app/model/migrations/migrations/to0002/MigrationToVersion0002.ts
BamMironov: I don't think it is worth it, at least for `processUserActions` because if `processVideoPhotoItems`/`processAudioFiles` fails there is no need to even call `processUserActions` because it basically does the same job `processAudioFiles`/`processAudioFiles` does but in the user actions

PR #757, opened 2024-05-03
M2-6301: NIMH Applet Scheduling is not working as expected

anq83:
### 📝 Description

🔗 [Jira Ticket M2-6301](https://mindlogger.atlassian.net/browse/M2-6301)

Changes include:

- Fixes how the mobile and server side save/keep/analyze the date-time moment of activity/flow completion. In case of autocompletion this moment should be equal to availableTo - the right part of the current time window based on the event. Previously, for autocompletion it was Date.now() and that value set to the dto that transfered to server, whereas mobile app considered availableTo instead.
- Fixes how the available/scheduled group evaluators filter activitiy+event items based on dateFrom, dateTo. Previously we had a bug - endDate handled wrong way - considered start of day (00:00:00) i.o. end of day (23:59:59)
- Fixes the progress sync service - as consequence of the 1st error in this list - the progress overwritren by the wrong endAt value that received from api.

### DIFF
--- a/src/widgets/survey/model/services/ConstructCompletionsService.ts
+++ b/src/widgets/survey/model/services/ConstructCompletionsService.ts
@@ -65,7 +65,7 @@ type ConstructForFinishInput = {
   isAutocompletion: boolean;
 };
 
-type CompletionType = 'intermediate' | 'finish';
+export type CompletionType = 'intermediate' | 'finish';
 
 export type ConstructInput = (
   | ConstructForIntermediateInput
--- a/src/widgets/survey/model/services/ConstructCompletionsService.ts
+++ b/src/widgets/survey/model/services/ConstructCompletionsService.ts
@@ -1,4 +1,5 @@
 import { QueryClient } from '@tanstack/react-query';
+import { addMilliseconds } from 'date-fns';
 
 import { StoreProgress } from '@app/abstract/lib';
 import { IPushToQueue } from '@app/entities/activity';
@@ -16,6 +17,7 @@ import {
   getEntityProgress,
   getNow,
   getTimezoneOffset,
+  isEntityExpired,
   Logger,
   MixEvents,
   MixProperties,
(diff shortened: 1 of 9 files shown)


src/widgets/survey/model/services/ConstructCompletionsService.ts
```diff
@@ -117,12 +140,30 @@ export class ConstructCompletionsService {
...
+    availableTo: number | null,
+    isAutocompletion: boolean,
+  ): number {
+    if (!isAutocompletion || !availableTo) {
+      return getNow().getTime();
+    }
+
+    return completionType === 'intermediate'
+      ? availableTo
+      : addMilliseconds(availableTo, DistinguishInterimAndFinishLag).getTime();
```
BamMironov, line 164: Please elaborate on why we need to add 1 millisecond for Admind and Backend to be able to sort it the right way. Is there a better alternative for it?

BamMironov (changes requested):
Looks good in general.Although, I am not sure about adding 1 millisecond "for better sorting". Sounds like this is the issue of Backend or Admin

src/widgets/survey/model/services/ConstructCompletionsService.ts
anq83: Alternative: investigate all related code on BE & Admin. We've already spend some time for investigation and for communication (~2h for every person: me, Ihor, Uladzimir) and current approach looks as optimal. But for ideal approach - investigation and possible changes on BE/Admin sides. And considering all these changes on deployment. I don't think it's worth it.

BamMironov: @anq83  thank you for the clarification!
Please add a "TODO" comment then. I saw you already added a comment to a variable declaration, so just add a "TODO" note explicitly.

anq83: added

BamMironov (approved):

PR #772, opened 2024-05-17
M2-5700, M2-6663, M2-6284: Upload with Timer issues

anq83:
### 📝 Description

🔗 [Jira Ticket M2-5700](https://mindlogger.atlassian.net/browse/M2-5700)
🔗 [Jira Ticket M2-6663](https://mindlogger.atlassian.net/browse/M2-6663)
🔗 [Jira Ticket M2-6284](https://mindlogger.atlassian.net/browse/M2-6284)

Changes include:

- When activity/flow completed by timer - interrupted activity answers considered and sent now
- When activity/flow completed by timer - endAt will be set to availableTo-1sec in case of overdue


src/shared/lib/timers/AppTimer.ts
BamMironov, line 1: `AppTimer` is a simple class with a single responsibility. It should not depend on any domain logic or has more responsibility. Please move the logic from here to the place where the AppTimer instances are used

src/shared/lib/timers/hooks/useAppTimer.ts
```diff
@@ -15,7 +17,10 @@ const useAppTimer = (appTimerConfig: AppTimerConfig) => {
   callbacksRef.current = { onFinish };
 
   useEffect(() => {
-    const onTimerEnd = () => callbacksRef.current.onFinish();
+    const onTimerEnd = (): TimerCallbackFeedback => {
+      callbacksRef.current.onFinish();
+      return 'no-feedback';
+    };
```
BamMironov, line 23: Why isn't type `void` enough?

src/abstract/lib/types/primitive.ts
```diff
@@ -16,3 +16,5 @@ export type AppletWithVersion = {
 };
 
 export type NoParamVoidFunction = () => void;
+
+export type TimerCallbackFeedback = 'no-feedback' | 'request-for-postpone';
```
BamMironov, line 20: Not sure what it means. Request for postponement of what?

src/features/pass-survey/model/hooks/useIdleTimer.ts
```diff
@@ -41,8 +42,9 @@ const useIdleTimer = (input: UseIdleTimerInput): UseIdleTimerResult => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  const onIdleElapsed = () => {
+  const onIdleElapsed = (): TimerCallbackFeedback => {
     onFinish();
+    return 'no-feedback';
```
BamMironov, line 47: Why isn't type `void` enough?

src/widgets/survey/model/hooks/useFlowStateActions.ts
```diff
@@ -128,11 +137,22 @@ export function useFlowStateActions({
     }
   }
 
-  function completeByTimer() {
+  function completeByTimer(): TimerCallbackFeedback {
```
BamMironov, line 140: I am not sure why it should return something related to timer callback when it is not defined as a callback (it is defined just as an exportable function) and has nothing to do with how it is going to be used.

src/widgets/survey/model/hooks/useFlowStateActions.ts
```diff
@@ -128,11 +137,22 @@ export function useFlowStateActions({
...
     if (isLastStep(record) || isSummaryStep(record)) {
-      return;
+      return 'no-feedback';
+    }
+
+    if (isInterimStep(record) && InterimSubmitObservable.processing) {
+      Logger.log(
+        `[useFlowStateActions.completeByTimer] Requested for postpone`,
+      );
+      return 'request-for-postpone';
```
BamMironov, line 155: Please move the logic out of the function. It should be able to do both: completeByTimer and check `isInterimStep(record) && InterimSubmitObservable.processing`

src/shared/lib/timers/AppTimer.ts
BamMironov: I can see that all of these can be moved to widgets/survey/model/hooks/useTimer. Also, we should probably rename this hook as has more responsibility now OR make `useTimer` a simple wrapper around AppTimer and create a domain hook that implements the logic you have described.

src/shared/lib/observables/interimSubmitObservable.ts
```diff
@@ -0,0 +1,31 @@
...
+  extends CommonObservable
+  implements IInterimSubmitObservable
+{
+  constructor() {
+    super();
+  }
+
+  private _processing: boolean = false;
+
+  public get processing() {
```
BamMironov, line 17: Please implement `IMutex` interface instead of `processing` getter and `reset` because it does exactly the same thing

BamMironov (changes requested):

src/shared/lib/timers/hooks/useAppTimer.ts
anq83: To make the code more robust and clear. In case of `void` is very simple to forget to return needed result.

src/abstract/lib/types/primitive.ts
anq83: Yes, callback implicitly asks timer to wait for a while and repead callback again.

src/features/pass-survey/model/hooks/useIdleTimer.ts
anq83: Answered above.

src/widgets/survey/model/hooks/useFlowStateActions.ts
anq83: Otherwise, we face with race-condition when callback is fired during uploading on Interim screen and then we switch to the Finish screen and doing the same actions (autocompletion as well).

src/widgets/survey/model/hooks/useFlowStateActions.ts
anq83: We already have similar logic in this handler. We return in case if we're on Summary on on Finish screen. Please specify.

src/shared/lib/observables/interimSubmitObservable.ts
anq83: Mutex doesn't help here. Where would you suggest to place it? 
We change observable state from the Interim screen. Should Mutex be there? I don't think so, as mutex usually bound with any instance of any class.

src/shared/lib/observables/interimSubmitObservable.ts
```diff
@@ -0,0 +1,31 @@
+import { CommonObservable } from '../utils';
+
+export interface IInterimSubmitObservable {
```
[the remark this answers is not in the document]
anq83: ok

src/shared/lib/timers/AppTimer.ts
anq83: I'll try an approah with inheritance

src/widgets/survey/model/hooks/useFlowStateActions.ts
BamMironov: The function `completeByTimer` has a single responsibility (to upsert the flow storage record with values) that should not be violated.

src/shared/lib/timers/hooks/useAppTimer.ts
BamMironov: It is not very clear why a callback should return a 'no-feedback' string though.

src/abstract/lib/types/primitive.ts
BamMironov: Maybe, it would be a good idea to call `postpone` operation when it is needed instead of relying on strings

src/shared/lib/timers/hooks/useAppTimer.ts
anq83: reverted

src/abstract/lib/types/primitive.ts
anq83: reverted

src/features/pass-survey/model/hooks/useIdleTimer.ts
anq83: reverted

src/widgets/survey/model/hooks/useFlowStateActions.ts
anq83: Resolved by Postponer

src/widgets/survey/model/hooks/useFlowStateActions.ts
anq83: Resolved by Postponer

src/shared/lib/observables/interimSubmitObservable.ts
anq83: 1st looks good.

PR #777, opened 2024-05-20
M2-6326: Removed auto-default value for TimeRange item

vmkhitaryanscn:
### 📝 Description

🔗 [Jira Ticket M2-6326](https://mindlogger.atlassian.net/browse/M2-6326)

Changes include:

- removing default date value for TimeRange item
- a fix for unit tests
- Changed time Range item contract

### DIFF
--- a/src/features/pass-survey/model/hooks/useActivityStepper.ts
+++ b/src/features/pass-survey/model/hooks/useActivityStepper.ts
@@ -38,14 +38,14 @@ function useActivityStepper(state: ActivityState | undefined) {
   const canSkip =
     !!currentPipelineItem?.isSkippable && !hasAnswer && !isSplashStep;
 
-  const canContinue =
+  const canMoveNext =
     isTutorialStep ||
     isMessageStep ||
     isAbTestStep ||
     currentPipelineItem?.isSkippable ||
-    (hasAnswer && (!additionalAnswerRequired || hasAdditionalAnswer));
-
-  const canMoveNext = canContinue && answerValidator.isValidAnswer();
+    (hasAnswer &&
+      (!additionalAnswerRequired || hasAdditionalAnswer) &&
+      answerValidator.isValidAnswer());
 
   const canMoveBack = currentPipelineItem?.isAbleToMoveBack;
   const canReset =
--- a/src/features/pass-survey/model/hooks/useActivityStepper.ts
+++ b/src/features/pass-survey/model/hooks/useActivityStepper.ts
@@ -2,6 +2,7 @@ import {
   ActivityItemType,
   ActivityState,
   onIncorrectAnswerGiven,
+  TimeRangeResponse,
 } from '../../lib';
 import AnswerValidator from '../AnswerValidator';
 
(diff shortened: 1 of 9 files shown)


src/features/pass-survey/model/hooks/useActivityStepper.ts
```diff
@@ -34,14 +36,23 @@ function useActivityStepper(state: ActivityState | undefined) {
     answers[step]?.additionalAnswer != null &&
     answers[step]?.additionalAnswer !== '';
 
+  const invalidTimeRangeAnswer =
+    isTimeRangeStep &&
+    ((answers[step]?.answer as TimeRangeResponse)?.startTime === null ||
+      (answers[step]?.answer as TimeRangeResponse)?.endTime === null);
```
BamMironov, line 42: The check is correct. However, I'd like to propose some refactoring. Here is my idea:
We have an `AnswerValidator` instance here. We could add a new method called `isValidAnswer` that could perform the check-ups described above for the Time Range item and just check for undefined | null for other items.
Then, we could have something like this:
`const canMoveNext = canContinue && answerValidator.isValidAnswer();`

src/widgets/survey/model/mappers.ts
```diff
@@ -211,19 +211,23 @@ function convertToDateAnswerAnswer(answer: Answer): AnswerDto {
 }
 
 function convertToTimeRangeAnswer(answer: Answer): AnswerDto {
+  const defaultDate = new Date();
+  const defaultHours = defaultDate.getHours();
+  const defaultMinutes = defaultDate.getMinutes();
```
BamMironov, line 216: Has this been approved by the BA team?

BamMironov (changes requested):
Please consider a little refactoring

src/features/pass-survey/model/hooks/useActivityStepper.ts
```diff
@@ -36,12 +37,16 @@ function useActivityStepper(state: ActivityState | undefined) {
...
-  const canMoveNext =
+
+  const canContinue =
     isTutorialStep ||
     isMessageStep ||
     isAbTestStep ||
     currentPipelineItem?.isSkippable ||
     (hasAnswer && (!additionalAnswerRequired || hasAdditionalAnswer));
+
+  const canMoveNext = canContinue && answerValidator.isValidAnswer();
```
BamMironov, line 48: I just realized that we should bypass items if the items are skippable. Please move the validation to the 46 line:
`(hasAnswer && (!additionalAnswerRequired || hasAdditionalAnswer)) && answerValidator.isValidAnswer();`

src/widgets/survey/model/mappers.ts
BamMironov: If a Time Range item is skipped it can have one of the fields selected. In my opinion, in this case, we should swap the value to a default one.
Example:
The user chose only the Time From and hit the "Skip" button. When it comes to submitting the data, the mapper should check if both fields are filled with values: if not then set the default values.

@moiskillnadne what would you say? How is that handled on the Web App?

BamMironov (changes requested):

src/widgets/survey/model/mappers.ts
moiskillnadne: @vmkhitaryanscn @BamMironov 
The web application doesn`t have the default values, so that means the web app doesn`t handle it.

If the user chooses only from time and clicks the next the following behavior will be:
The item answer for the to-time field will be empty
The answers mapper leaves this as is

BamMironov (approved):

PR #790, opened 2024-06-04
M2-6652: [Mobile] Single select options and tooltips

anq83:
### 📝 Description

🔗 [Jira Ticket M2-6652](https://mindlogger.atlassian.net/browse/M2-6652)

Changes include:

- Fix some attributes to allow aqa continue some tests

### DIFF
--- a/src/shared/ui/survey/CheckBox/CheckBox.item.tsx
+++ b/src/shared/ui/survey/CheckBox/CheckBox.item.tsx
@@ -71,7 +71,7 @@ const CheckBoxItem: FC<Props> = ({
             {!!tooltip && (
               <Tooltip
                 markdown={tooltipText}
-                accessibilityLabel="checkbox-tooltip-view"
+                accessibilityLabel={`checkbox-tooltip-view-${tooltipText}`}
               >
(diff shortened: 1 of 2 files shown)


src/shared/ui/survey/CheckBox/CheckBox.item.tsx
```diff
@@ -71,7 +71,7 @@ const CheckBoxItem: FC<Props> = ({
             {!!tooltip && (
               <Tooltip
                 markdown={tooltipText}
-                accessibilityLabel="checkbox-tooltip-view"
+                accessibilityLabel={`checkbox-tooltip-view-${tooltipText}`}
```
BamMironov, line 74: Please make sure that nothing breaks if `tooltipText` contains markdown

BamMironov (approved):
Looks good but double-check if markdown works in accessibility label

src/shared/ui/survey/CheckBox/CheckBox.item.tsx
anq83: it's ok, tried many special char-s

PR #813, opened 2024-07-08
M2-7248, M2-7246: bugfixes for autocompletion delivery part 2

anq83:
### 📝 Description

🔗 [Jira Ticket M2-7248](https://mindlogger.atlassian.net/browse/M2-7248)
🔗 [Jira Ticket M2-7246](https://mindlogger.atlassian.net/browse/M2-7246)

Changes include:

- Fix for iOS: swipe from up to down - previousy -  the event 'active' registered
- Fix for restart-resume, when alert opened for a long time, the entity could expire
- Lite refactoring in RootNavigator: triggers for autocompletion when: go-to-foreground, on-app-start, go-to-online
- Fix for isQueueEmpty when autocompletion with upload works, to make decision if blocker screens should be opened

### DIFF
--- a/src/shared/lib/hooks/useOnForegroundDebounced.ts
+++ b/src/shared/lib/hooks/useOnForegroundDebounced.ts
@@ -3,29 +3,32 @@ import { AppState, AppStateStatus } from 'react-native';
 
 import { useDebouncedCallback } from 'use-debounce';
 
+import { IS_IOS, Logger } from '../';
+
 type Options = {
   enabled: boolean;
 };
 
-const DebounceInterval = 100;
+const DebounceInterval = 200;
 
 function useOnForegroundDebounced(
   callback: () => void,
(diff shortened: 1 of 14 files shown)


src/shared/lib/hooks/useOnForegroundDebounced.ts
```diff
@@ -0,0 +1,43 @@
...
+  enabled: boolean;
+};
+
+const DebounceInterval = 100;
+
+function useOnForegroundDebounced(
+  callback: () => void,
+  options?: Partial<Options>,
+) {
+  const callbackRef = useRef(callback);
```
BamMironov, line 16: No need to use `useRef` for the callback here as `useDebouncedCallback` already does it under the hood

BamMironov (changes requested):

BamMironov (approved):

PR #831, opened 2024-07-24
M2-7321, M2-7433: [Mobile]: Handle tap on notification while autocompletion is working

anq83:
### 📝 Description

🔗 [Jira Ticket M2-7321](https://mindlogger.atlassian.net/browse/M2-7321)
🔗 [Jira Ticket M2-7433](https://mindlogger.atlassian.net/browse/M2-7433)

Changes include:

- Delayed notification
- Pink banner minor changes

### DIFF
--- a/src/features/tap-on-notification/model/services/NotificationPostponer.ts
+++ b/src/features/tap-on-notification/model/services/NotificationPostponer.ts
@@ -1,0 +1,67 @@
+import { Logger } from '@app/shared/lib';
+
+const PostponeDuration = 1000;
+
+class NotificationPostponer {
+  private timeoutId: NodeJS.Timeout | null;
+
+  private _getCurrentRoute: (() => string | undefined) | null = null;
+
+  private _action: (() => void) | null = null;
+
+  constructor() {
+    this.timeoutId = null;
+  }
+
+  private shouldBePostponed(): boolean {
+    const isAutocompletionWorking =
+      this._getCurrentRoute!() === 'Autocompletion';
+
+    return isAutocompletionWorking;
+  }
+
+  private postpone() {
+    this.timeoutId = setTimeout(() => {
+      this.try();
+    }, PostponeDuration);
+  }
+
+  private try(): boolean {
+    if (this.shouldBePostponed()) {
+      this.postpone();
+
+      return false;
+    } else {
+      this._action!();
+
+      return true;
+    }
+  }
+
+  public set getCurrentRoute(value: () => string | undefined) {
+    this._getCurrentRoute = value;
+  }
+
+  public set action(value: () => void) {
+    this._action = value;
+  }
+
+  public tryExecute(): boolean {
+    const success = this.try();
+
+    if (!success) {
+      Logger.log('[NotificationPostponer.try] Postponed');
+    }
+
+    return success;
+  }
+
+  public reset() {
+    if (this.timeoutId) {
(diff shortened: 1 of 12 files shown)


src/features/tap-on-notification/model/services/NotificationPostponer.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  public reset() {
+    if (this.timeoutId) {
+      clearTimeout(this.timeoutId);
+    }
+    this._action = null;
+  }
+}
+
+export default NotificationPostponer;
```
BamMironov, line 67: The class itself is okay. However, the same functionality is implemented within `InterimInActionPostponer` class.
I would propose refactoring `InterimInActionPostponer` and then `NotificationPostponer` so that they

- either extend `ActionPostponer` class that contains `shouldBePostponed`, `tryExecute`, `try` and `reset` methods.
- or implements `IActionPostponer` interface that describes aforementioned methods.
I wouldn't add `_getCurrentRoute`/`getCurrentRoute` and `_action`/`action` methods as they don't relate to the class responsibility, hence, should be called outside of the class.

BamMironov (changes requested):

src/features/tap-on-notification/model/services/NotificationPostponer.ts
anq83: 1. Please clarify: what do you mean under **extend ActionPostponer** ? To create super class ActionPostponer and 2 subclasses?
2. We cannot create interface that contain private or protected methods. re. getCurrentRoute, action - this is up to concrete postponer how to implement shouldBePostponed check. Interim postponer uses another check based on InterimSubmitMutex.

BamMironov: Here is the corrected and formatted text:

1. Yes, `ActionPostponer` could be a parent class from which `NotificationPostponer` and `InterimInActionPostponer` could inherit, for instance:
    ```javascript
    class NotificationPostponer extends ActionPostponer {
    ```

2. Sure, we can't; I missed this. However, I would like to emphasize the class responsibility: to postpone work. `getCurrentRoute` and `actions` look irrelevant to the problem the class solves. What can we do about that?

    - Since `getCurrentRoute` is called only inside the `shouldBePostponed` method, this method should serve as a connection between the outer world and the class. Therefore, make `shouldBePostponed` a function that could be passed as an `ActionPostponer` constructor parameter:
        ```javascript
        class ActionPostponer {
          constructor(shouldBePostponed: () => boolean) {
        ```
    - Next, `action` could also be moved to the constructor of the class for the same reason mentioned above. Moreover, if you look at the `InterimInActionPostponer` implementation, you will see that it's already been implemented this way. The only thing to do is to move it to the parent `ActionPostponer` class:
        ```javascript
        class ActionPostponer {
          constructor(shouldBePostponed: () => boolean, actionToPostpone: () => void) {
        ```
    - Or even make it accept an object of functions to make things clearer and more robust:
        ```javascript
        type ActionPostponerParams = {
          shouldBePostponed: () => boolean,
          actionToPostpone: () => void
        }

        class ActionPostponer {
          constructor(params: ActionPostponerParams) {
        ```

anq83: implemented 1st

BamMironov (approved):

PR #834, opened 2024-07-26
M2-7464: [Mobile] Fix DatePickerItem unit tests

anq83:
### 📝 Description

After 
https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/833
I've got error (I'm under GMT+2/+3, ~12:30 PM)

image

🔗 [Jira Ticket M2-7464](https://mindlogger.atlassian.net/browse/M2-7464)

Changes include:

- Modified useless unit test to get practical sence
- Added another unit test to the same component

The test itself was wrong:
it("Should render new Date(0) when value is 1970-01-01", ... )
new Date(0) doesn't equal to new Date(1970,0,1) because the 1st is based on UTC time, the 2nd - on local.
So, I've changed it to:
it('Should consume new Date(1970, 0, 1) when props value is "1970-01-01"', ... )

### 📸 Screenshots

Some information:

image

image

### DIFF
--- a/src/shared/ui/survey/tests/DatePickerItem.test.tsx
+++ b/src/shared/ui/survey/tests/DatePickerItem.test.tsx
@@ -21,7 +21,7 @@ describe('Test DatePickerItem', () => {
     expect(placeholder).toBe(expected);
   });
 
-  it('Should render new Date(0) when value is 1970-01-01', () => {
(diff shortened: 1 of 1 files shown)


ChaconC (approved):
Confirmed tests pass with different locale TZ

PR #845, opened 2024-08-13
M2-6096, M2-2872: Fixes the big black square in Flanker's summary view

BamMironov:
### 📝 Description

🔗 [Jira Ticket M2-2872](https://mindlogger.atlassian.net/browse/M2-2872)
🔗 [Jira Ticket M2-6096](https://mindlogger.atlassian.net/browse/M2-6096)

Changes include:

- Made Flanker work offline by providing the Swift module with access to the real cache directory where the cached images are stored.
- Set the default background color of the `textLabel` view to transparent (aka `.clear`). Supposedly, updating Xcode led to updates in the iOS SDK or changes in the underlying UIKit framework that came with the version of Xcode. By default it used to be `nil` now it is apparently `"black"`
- Migration step. This makes sure to prefetch and cache Flanker’s images for the users that already have Flanker activity in their app and just want to update the app instead of a fresh install.
 
### 📸 Screenshots

| Before                      |
| -------------------------------------- |
|  |

| After                                 |
| ------------------------------------- |
| |


ios/FlankerNativeComponents/ImageCacheManager.swift
```diff
@@ -0,0 +1,50 @@
...
+    }
+    
+    return UIImage(contentsOfFile: filePath)
+  }
+     
+  public static func loadCachedImages(from urls: [String]) -> [UIImage?] {
+    return urls.compactMap { url in
+      loadCachedImage(from: url)
+    }
+  }
```
ChaconC, line 49: I believe this method is not used, are there plans for using this in the future?

ChaconC (dismissed):
Looking good, tested on device and images are cached and retrieved correct when offline. Also the ResultView displays the correct color. 
Minor comment but good to pre-approve 👍

ios/FlankerNativeComponents/ImageCacheManager.swift
BamMironov: You are correct. At first, I intended to retrieve all cached images at once but then I decided to do it one by one to minimize the `ImageLoader` class changes. Ok, let's delete this.
I wish there is a way for Xcode to highlight unused methods.

BamMironov:
@ChaconC  @felipeMetaLab  please review the app once again as I added a commit regarding data migration.

In short, this migration makes sure to prefetch and cache Flanker’s images for the users that already have Flanker activity in their app and just want to update the app instead of a fresh install.

ChaconC (approved):
all good, tested migrations locally 🚀

felipeMetaLab:
Every thing is looking good. :D

PR #846, opened 2024-08-21
M2-7508: Image resize

felipeMetaLab:
### 📝 Description

🔗 [Jira Ticket M2-#](https://mindlogger.atlassian.net/browse/M2-7508)

This PR contemplates image resize, its handling with image even if the size of image is not passed.
that is, it is now possible to pass a size to the image with "=widthxheight" or refrain from passing it , so it will catch the original  size from the image.
The resize contemplate aspect ratio when width exceeds the view port width.

Changes include:

- new regex to process image link
- new rule for markdown image

### 📸 Screenshots
Setting up the in web admin
Screenshot 2024-08-20 at 10 43 15 PM

Result:
Screenshot 2024-08-20 at 10 43 34 PM

no Internet ( I tested removing internet from simulator and turning off my wi-fi and reloading the app using terminal pressing"r"):
Screenshot 2024-08-20 at 11 21 15 PM

### 🪤 Peer Testing
Assuming your environment is configured, you must:
1 - register the images on admin app.
2 - change the size like: 

### ✏️ Notes

It is expected that there will be differences between the admin and the app such as spacing, centering, softbreak and other style differences regarding markdown.

the resize contemplate aspect ratio when width exceeds the view port width.


src/shared/lib/markDownRules.tsx
```diff
@@ -547,7 +575,52 @@ const getContainerAlignTag = (parents: ASTNode[]): AlignmentTag | undefined => {
...
+
+  const getImageSize = (url: string) => {
+    return new Promise<{ width: number; height: number }>((resolve, reject) => {
+      Image.getSize(
+        url,
+        (width, height) => resolve({ width, height }),
+        error => reject(error),
+      );
+    });
+  };
```
BamMironov, line 589: Please move the `getImageSize` function outside of the `preprocessImageLinks` scope and make it exportable. So that we could cover it with tests later (or now?).

Also, if the `error` is not used anywhere in the code it might be a good idea to just resolve the promise with `null`. Otherwise, the code that utilizes `getImageSize` should handle the error, e.g. put it in the logs

src/shared/lib/markDownRules.tsx
```diff
@@ -547,7 +575,52 @@ const getContainerAlignTag = (parents: ASTNode[]): AlignmentTag | undefined => {
...
+
+  const matches = [...content.matchAll(regexImage)];
+
+  const promises = matches.map(match => {
```
BamMironov, line 593: I think you could drastically improve readability of the peace of code by re-writing the `preprocessImageLinks` to async/await  style instead of promises.
This way, you will get rid of the `promises ` variable (and the subsequent call of `Promise.all`) and `Promise.resolve` calls

src/shared/lib/markDownRules.tsx
```diff
@@ -547,7 +575,52 @@ const getContainerAlignTag = (parents: ASTNode[]): AlignmentTag | undefined => {
...
+      content = content.replace(fullMatch, updatedMatch);
+      return Promise.resolve();
+    } else {
+      return getImageSize(url)
+        .then(({ width, height }) => {
+          const newUrl = `${url}?width=${width}&height=${height}`;
+          const updatedMatch = fullMatch.replace(url, newUrl);
+          content = content.replace(fullMatch, updatedMatch);
+        })
+        .catch(_error => {
```
BamMironov, line 611: Please handle the `error` somehow. For instance, you could put it in the logs

src/shared/ui/MarkdownView.tsx
```diff
@@ -32,14 +32,22 @@ declare module 'react-native-markdown-display' {
 }
 
 const MarkdownView: FC<Props> = ({ content, markdownStyle, rules }) => {
+  const [processedContent, setProcessedContent] = useState<string>('');
+
+  useEffect(() => {
+    preprocessImageLinks(content).then(result => {
+      setProcessedContent(result);
+    });
+  }, [content]);
```
BamMironov, line 41: `preprocessImageLinks` utilizes [Image.getSize](https://reactnative.dev/docs/image#getsize) that **downloads** the images first and then calculates the size. 
What if there is no Internet connection? Or, what if the Internet Connection is poor? I believe `processedContent` could be empty for a while in this case

BamMironov (changes requested):
IMHO, although the solution is working at the moment, I don't think it's a reliable one.
My three main concerns are:
1. The `preprocessImageLinks` function blocks the content rendering (see the comments below regarding `Image.getSize` function). As we support Offline Mode we should make sure everything works smoothly, without interruptions.
2. Both `preprocessImageLinks` function AND markDownRules's `image` rule are literally doing one job - to parse images meta data. Meaning, that the `image` rule can not function properly without `preprocessImageLinks`'s job being done previously. This is an **implicit dependency** that will lead to unexpected bugs and is hard to maintain. 
3. The lack of error handling when working with regexes. 

However, I believe that most of the code here can be easily used in a plugin solution for markdown-it library via `.use(plugin)`: https://github.com/markdown-it/markdown-it?tab=readme-ov-file#plugins-load 
so that the plugin will have a single responsibility of parsing images in one dedicated place.

felipeMetaLab:
@BamMironov  the problem in use .use(plugin):
The bottleneck of this solution is that we are receiving with url "=widthxheight", this way the plugin is not able to identify the url as image but as text.
We could handle with preprocessImageLinks, to transform and forcing to appear in plugin (but do you agree we would have 2 points handling with same thing "Image" ? maybe its not the best stuff), but it become extremely hard to handle with asynchronism,  making harder to maintainability.

BamMironov:
@felipeMetaLab  Please consider changes in the "feat/M2-7508_plugin-solution" branch.
There, you can find the patched version of [markdown-it-imsize](https://github.com/tatsy/markdown-it-imsize) that now works well in React Native. The same library is used in Admin Panel.
It offers a very convenient way of retrieving the `width` and `height` for images when a URL contains a " =widthxheight" postfix. You can find them in the `attributes` property of the `node`:
image
image

Also, `getImageSize` was edited:
- The usage of [Image.getSize](https://reactnative.dev/docs/image#getsize) has been removed
- Add `ImageDimensions` native module to calculate images dimensions with `getImageDimensionsSync` method.
- The function is now synchronous

`preprocessImageLinks` is deprecated. You will want to edit `getSizeByURLQueryParams` and use it with `getImageSize` to figure out the exact `width` and `height` considering the image size aspect ratio.

ChaconC:
Can we close this PR @felipeMetaLab ?

PR #851, opened 2024-08-26
M2-7457:  Character Counter

felipeMetaLab:
### 📝 Description
🔗 [Jira Ticket M2-#](https://mindlogger.atlassian.net/browse/M2-7457)
Creating CharacterCounter component and adding it to Paragraph text

Changes include:

- New component.
- implemented on paragraph

### 📸 Screenshots
Simulator Screenshot - mindLogger - 2024-08-26 at 14 46 32
Simulator Screenshot - mindLogger - 2024-08-26 at 14 46 27
Simulator Screenshot - mindLogger - 2024-08-27 at 11 02 28

### 🪤 Peer Testing
1 - crating a paragraph text
2 - the counter must appears on bottom right
3 - validate the color behavios when onFocus and onBlur
4 - validate the number of typed character and limit


src/shared/ui/survey/CharacterCounter.tsx
```diff
@@ -0,0 +1,27 @@
...
+  limit,
+  fontSize = 14,
+  focused = false,
+}) => {
+  return (
+    <CharacterCounterText
+      color={focused ? 'primary' : 'grey4'}
+      fontSize={fontSize}
+    >
+      {numberOfCharacters}/{limit} characters
```
ChaconC, line 22: Use translatable strings, example: `i18n.t('text_entry:character_count', { count, limit })`

src/shared/ui/survey/ParagraphText.tsx
```diff
@@ -1,11 +1,13 @@
-import React, { FC } from 'react';
+import React, { FC, useState } from 'react';
 import { StyleSheet, TextInputProps, View } from 'react-native';
 
 import { useTranslation } from 'react-i18next';
 
 import { colors } from '@shared/lib';
 import { LongTextInput } from '@shared/ui';
 
+import CharacterCounter from './CharacterCounter';
```
ChaconC, line 9: import via `@shared/ui`

src/shared/ui/survey/CharacterCounter.tsx
ChaconC, line 1: `CharacterCounter` is not an item type element, it shouldn't exist at the `src/shared/ui/survey` level.
Other similar elements, for example `BackButton` or `Input`, go in `src/shared/ui`

src/shared/ui/CharacterCounterText.tsx
```diff
@@ -0,0 +1,42 @@
...
+        color: colors[color] || color,
+      }),
+    },
+  },
+  {
+    inlineProps: new Set(['fontSize', 'color']),
+  },
+);
+
+export default focusableInputHOC(CharacterCounterText);
```
ChaconC, line 42: why is this a `InputHOC`? This should be a label, as in not focusable

src/shared/ui/survey/tests/CharacterCounter.test.tsx
ChaconC, line 1: Same here, should import from `@shared/ui`, this test should probably go in `src/shared/ui/tests`

ChaconC (changes requested):

src/shared/ui/survey/CharacterCounter.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { FC } from 'react';
+
+import CharacterCounterText from '../CharacterCounterText';
+type Props = {
```
ChaconC, line 4: add space between import and declarations

src/shared/ui/CharacterCounter.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <Text
+      style={[
+        styles.CharacterCounterText,
+        {
+          fontSize,
+          color: focused ? colors.primary : colors.grey4,
+        },
+      ]}
```
BamMironov, line 33: Please avoid using inline styles for Tamagui's components unless you have a good reason

src/shared/ui/survey/ParagraphText.tsx
```diff
@@ -15,10 +15,13 @@ type Props = {
 } & Omit<TextInputProps, 'value' | 'onChange'>;
 
 const ParagraphText: FC<Props> = ({ value, onChange, config, ...props }) => {
+  const [characterCounter, setCharacterCounter] = useState(0);
```
BamMironov, line 18: I think you could use `value.length` instead

src/shared/ui/CharacterCounter.tsx
BamMironov, line 1: In my opinion, it would be great to add validation against props here.
For instance, `limit` should not be a zero. Or `numberOfCharacters` should not exceed `limit`. 
If any of the check above fail then may be we should throw `Error` here, in the `CharacterCounter`'s render.

On the other hand, the validation can be performed somewhere else, e.g. in the parent component. However, the name "CharacterCounter" suggests us that it's the responsibility of the component to validate the incoming props.

BamMironov (changes requested):
Please consider a couple of comments from my side. But generally, it looks good to me

src/shared/ui/CharacterCounter.tsx
```diff
@@ -0,0 +1,48 @@
...
+    >
+      {numberOfCharacters}/{limit} {t('character_counter:characters')}
```
ChaconC, line 35: `{numberOfCharacters}/{limit}` should be included as parameters of the string, as not all locales will follow the same "0/50 characters" convention

src/shared/ui/CharacterCounter.tsx
BamMironov: Yeah, right. If the system should allow users to enter more characters than the `limit` value and highlight an error then the _numberOfCharacters <= limit_ validation should be omitted.

felipeMetaLab: admin doesn't allow saving 0 or negative numbers

BamMironov: If the efforts to throw an error are significant then of course it doesn't make sense.

Generally speaking, an independent system should not rely on the stability of the external system: the Admin Panel code has nothing to do with the Mobile App code. If something changes/breaks on the Admin Panel, the Mobile App won't know about it.

ChaconC:
@felipeMetaLab can you review the [failing test cases](https://github.com/ChildMindInstitute/mindlogger-app-refactor/actions/runs/10584356662/job/29328426256?pr=851)

src/shared/ui/CharacterCounter.tsx
felipeMetaLab: I added some validations

src/shared/ui/CharacterCounter.tsx
```diff
@@ -0,0 +1,64 @@
...
+    fontWeight: '400',
+    fontSize: 14,
+  },
+  focusedColor: {
+    color: colors.primary,
+  },
+  unfocusedColor: {
+    color: colors.grey4,
+  },
+  WarnColor: {
```
ChaconC, line 59: `warnColor`

ChaconC (approved):
Ran tests locally and all is passing.
I'm fine as far as my comments, but curious for @BamMironov to take a final pass.

felipeMetaLab:
@BamMironov  I added some validations,  when you have a chance, can you validate again and check if every thing is good for you ? 
Thanks :)

BamMironov:
@felipeMetaLab The code looks great!

However, when I ran the app from your git branch to test the component I noticed that I was not able to enter more symbols than `limit`.


https://github.com/user-attachments/assets/422e31dc-5f84-4255-817f-873c06876b7f



According to **Acceptance Criteria 1** from [M2-7457](https://mindlogger.atlassian.net/browse/M2-7457):

**Given**: Respondent accesses Paragraph Text Item

**When**: The inputted text reaches the max characters limit set in the Admin Panel

**Then**: The respondent is still capable to enter any more symbols/ text to overflow the max character limit

**AND**
The text box enters into an error state, and respondent cannot move to next item

felipeMetaLab:
@BamMironov that's right, it is the normal behavior of paragraph, the fact is that the component "character counter" accepts more then the limit, we can force some fake limit in paragraph to test the character counter like the image I sent.

BamMironov:
@felipeMetaLab I am sorry but it seems I don't understand the point you have made. Please clarify: so you're saying that **Acceptance Criteria 1** is no longer relevant and you won't fix the current behaviour?

BamMironov:
@felipeMetaLab  I see, thanks! Okay, so if maxLength is relevant only for tablets please make sure to specify that in the ticket so that the QA team is aware.

felipeMetaLab:
No you are right @BamMironov  I was look at the criteria of the other paragraph ticket,  it had a limit implicit, but I noticed it changed, sorry.

BamMironov (approved):

src/shared/ui/LongTextInput.tsx
```diff
@@ -40,7 +40,7 @@ export const LongTextInput = styled(
 
     focusable: true,
     multiline: true,
-    scrollEnabled: true,
+    scrollEnabled: !!isTablet(),
```
ChaconC, line 43: Unnecessary double negation

src/features/pass-survey/model/AnswerValidator.ts
```diff
@@ -106,6 +106,16 @@ function AnswerValidator(params?: AnswerValidatorArgs): IAnswerValidator {
 
           return true;
         }
+        case 'ParagraphText': {
+          const numberOfCharacter = `${currentAnswer?.answer}`.length;
```
ChaconC, line 110: Add `answer` cast to `ParagraphTextResponse`
```suggestion
          const answer = currentAnswer?.answer as ParagraphTextResponse;
          const limit = currentPipelineItem.payload.maxLength;

          if (limit < answer.length) {
            return false;
          }
```

src/features/pass-survey/model/AnswerValidator.ts
```diff
@@ -106,6 +106,16 @@ function AnswerValidator(params?: AnswerValidatorArgs): IAnswerValidator {
...
+          const limit = currentPipelineItem.payload.maxLength;
+
+          if (limit < numberOfCharacter) {
+            return false;
+          }
```
ChaconC, line 115: Cast `answer` to `ParagraphTextResponse`, fits better with the established pattern in the rest of the validations.

```suggestion
          const answer = currentAnswer?.answer as ParagraphTextResponse;
          const limit = currentPipelineItem.payload.maxLength;

          if (limit < answer.length) {
            return false;
          }
```

PR #853, opened 2024-09-04
M2-7508: New resize image with plugin

felipeMetaLab:
### 📝 Description

🔗 [Jira Ticket M2-7508](https://mindlogger.atlassian.net/browse/M2-7508)

This PR contemplates image resize, its handling with image even if the size of image is not passed.
that is, it is now possible to pass a size to the image with "=widthxheight" or refrain from passing it , so it will catch the original  size from the image.
The resize contemplate aspect ratio when width exceeds the view port width.

Changes include:

- new regex to process image link
- new rule for markdown image

### 📸 Screenshots

Result:
IOS:
Screenshot 2024-09-04 at 12 56 48 PM
Screenshot 2024-09-04 at 12 57 18 PM

no Internet ( I tested removing internet from simulator and turning off my wi-fi and reloading the app using terminal pressing"r"):
Screenshot 2024-08-20 at 11 21 15 PM

Android:
Screenshot 2024-09-05 at 11 26 45 AM

### 🪤 Peer Testing
Assuming your environment is configured, you must:
1 - register the images on admin app.
2 - change the size like: 

### ✏️ Notes

It is expected that there will be differences between the admin and the app such as spacing, centering, softbreak and other style differences regarding markdown.

the resize contemplate aspect ratio when width exceeds the view port width.


ChaconC:
Please make sure to rebase your branch to include the most recent commits from the target branch `release/2.2.2`

src/shared/lib/markDownRules.tsx
```diff
@@ -546,8 +560,44 @@ const getContainerAlignTag = (parents: ASTNode[]): AlignmentTag | undefined => {
   return tag;
 };
 
-export const preprocessImageLinks = (content: string) => {
-  return content?.replace(/(!\[.*\]\s*\(.*?) =\d*x\d*(\))/g, '$1$2');
-};
+export const preprocessImageLinks = (content: string): string => {
```
ChaconC, line 563: I'm still generally against the notion of modifying the url, and would instead prefer to use markdown-it-imsize to help determine dimensions without modifying content. Since images are cached I believe this can be handled in the image node instead without major changes in your code.

src/shared/lib/markDownRules.tsx
```diff
@@ -418,12 +431,28 @@ const markDownRules: RenderRules = {
...
     const isCached = !!CacheManager.entries[node.attributes.src];
 
+    if (!imageSize.width || !imageSize.height) {
+      const imageDimensions = getImageSize(src);
+
+      if (!imageDimensions) {
+        throw Error(
+          `[markDownRules.image]: No image dimension found for url: ${src}`,
+        );
+      }
```
ChaconC, line 441: Handling this error should be part of `getImageSize`, also should this instead fail gracefully and let MarkdownView render the image as is?

src/shared/lib/markDownRules.tsx
```diff
@@ -418,12 +431,28 @@ const markDownRules: RenderRules = {
...
+
+      const { width, height } = imageDimensions;
+      imageSize = {
+        width: Number(width),
+        height: Number(height),
+      };
```
ChaconC, line 447: `getImageSize` already returns a `number`, you could instead assign the return value directly

ChaconC (approved):
Please make sure to squash-merge and update title

PR #856, opened 2024-09-10
fix: local development patches in Podfile

qiushihe:
### 📝 Description

This PR updates `Podfile` with some patches needed for local development:

* Remove `.xcode.env.local` to avoid an issue with `node` path during compliation
* Patch `YogaKit`'s `YGLayout.m`
* Patch `Flipper`'s `FlipperTransportTypes.h`

### 📸 Screenshots

N/A

### 🪤 Peer Testing

1. Ensure the repo and Xcode is in a clean state:
```
$ rm -fr node_modules
$ rm -fr ~/Library/Developer/Xcode/DerivedData/*
$ rm -fr ./ios/Pods
$ yarn clean-ios
```
2. Install node dependencies:
```
$ yarn install
```
3. Install pods without local development patches:
```
$ yarn pods
```
4. Build and run the iOS app:
```
$ yarn ios
```
... and **this should fail**.
5. Clean up pods and Xcode:
```
$ rm -fr ~/Library/Developer/Xcode/DerivedData/*
$ rm -fr ./ios/Pods
$ yarn clean-ios
```
6. Install pods with local development patches:
```
$ LOCAL_DEV_PATCHES=true yarn pods
```
7. Build and run the iOS app again:
```
$ yarn ios
```
... and **this should not fail**.

### ✏️ Notes

N/A


ios/Podfile.lock
ChaconC, line 1: No need to commit this file

ios/Podfile
```diff
@@ -87,5 +87,16 @@ abstract_target 'MindloggerMobileCommonPods' do
...
     )
+
+    if ENV['LOCAL_DEV_PATCHES'] == 'true'
+      puts '[Local Dev] Removing .xcode.env.local'
+      %x(rm -f #{Pod::Config.instance.installation_root}/.xcode.env.local)
+
+      puts '[Local Dev] Patching YogaKit to fix YGMeasureView call signature'
+      %x(patch #{Pod::Config.instance.installation_root}/Pods/YogaKit/YogaKit/Source/YGLayout.m -N < #{Pod::Config.instance.installation_root}/patches/YGLayout.patch)
+
+      puts '[Local Dev] Patching YogaKit to fix FlipperTransportTypes'
```
ChaconC, line 98: Granted this was my mistake, this should read “Patching Flipper” instead of YogaKit

qiushihe:
@ChaconC okay I made the changes. Please take another look. Thanks!

farmerpaul (approved):
Oddly enough, when I ran `yarn pods` _without_ the `LOCAL_DEV_PATCHES=true` env var set, it succeeded fine! So I didn't even have to use the patches to get going.

For those who do, however, might be nice to add a convenient script to `package.json`:
```json
    "pods:local": "LOCAL_DEV_PATCHES=true bundle exec pod install --project-directory=ios",
```

ios/Podfile.lock
farmerpaul: Can you explain why we're not committing `Podfile.lock`? When I followed the instructions here, my `PODFILE CHECKSUM` line changed. Is that specific to my local environment, and committing it will cause issues for other engineers? Or do we have to make a habit of avoiding committing changes to `Podfile.lock` whenever it happens to change?

farmerpaul: Also my `project.pbxproj` file changes whenever I run `yarn pods`, in two places with the same change:
CleanShot 2024-09-11 at 09 14 57@2x

Maybe this is just another instance of just local-env-specific nuance that we'll have to deal with avoiding committing to the repo regularly?

ChaconC: This was my mistake I guess, the checksum of the `Podfile` changed with the addition of the patch files.

In general we're trying to avoid committing changes to Podfile.lock that are a result of indirect changes, for example by using different versions. For now we're sticking with version 1.14.3 so if your local cocoapods version differs it results in a Podfile.lock change. That's what I originally thought had happened

ChaconC:
> Oddly enough, when I ran `yarn pods` _without_ the `LOCAL_DEV_PATCHES=true` env var set, it succeeded fine! So I didn't even have to use the patches to get going.
> 
> For those who do, however, might be nice to add a convenient script to `package.json`:
> 
> ```json
>     "pods:local": "LOCAL_DEV_PATCHES=true bundle exec pod install --project-directory=ios",
> ```

It's supposed to work fine in both cases, difference being patches are only applied when the env var is set.

| LOCAL_DEV_PATCHES=true | LOCAL_DEV_PATCHES=false |
|-|-|
|CleanShot 2024-09-11 at 11 58 22@2x| CleanShot 2024-09-11 at 11 58 34@2x|


That said, I like your suggestion of adding a new run script.

farmerpaul:
> It's supposed to work fine in both cases, difference being patches are only applied when the env var is set.
Oh I misspoke – `yarn pods` worked both times, but also the first time I ran `yarn ios`, it also built and ran fine – right after running `yarn pods` without the env var set.

ChaconC (approved):
All good @qiushihe, thanks for taking care of this!

I think we're ok to add the `Podfile.lock` diff to this PR, since the checksum is likely related to the addition of the patches to Podfile.

PR #858, opened 2024-09-16
Chore/M2-7771 improve encrypt and removing wait from use upload progress

felipeMetaLab:
### 📝 Description

🔗 [Jira Ticket M2-7771](https://mindlogger.atlassian.net/browse/M2-7771)
in encrypt function  it was constantly using Buffer.concat.
By modifying  the function to collect encrypted chunks in an array and concatenating them once at the end we can improve the performance.
Removing the wait, we can improve even more the time.

Changes include:

- removing wait from the useUploadProgress
- improving encrypt function

### 🪤 Peer Testing

As long you already have the app properly configured, you must create an applet and test if the save flow is working properly and the time reduced for at least 1 second.

### ✏️ Notes
Before, the upload answers was taking about 1400 to 1600ms and encrypt was taking 1578 to 1650ms

### DIFF
--- a/src/shared/lib/hooks/useUploadProgress.ts
+++ b/src/shared/lib/hooks/useUploadProgress.ts
@@ -2,7 +2,7 @@ import { useEffect } from 'react';
 
 import { useTranslation } from 'react-i18next';
 
-import { UploadProgressObservable, wait } from '../';
+import { UploadProgressObservable } from '../';
 import { SecondLevelStep, UploadProgress } from '../observables/';
 
 import { useForceUpdate } from './';
@@ -100,7 +100,6 @@ const useUploadProgress = (): UseUploadProgressResult => {
   useEffect(() => {
     const onProgressChange = async (delay: number) => {
       update();
-      await wait(delay);
     };
 
     UploadProgressObservable.addObserver(onProgressChange);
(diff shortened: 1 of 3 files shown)


src/shared/lib/hooks/useUploadProgress.ts
```diff
@@ -100,7 +100,6 @@ const useUploadProgress = (): UseUploadProgressResult => {
   useEffect(() => {
     const onProgressChange = async (delay: number) => {
       update();
-      await wait(delay);
```
ChaconC, line 103: You can remove `async` and `delay: number` since it's no longer used

ChaconC (changes requested):
Can you share how you tested your results? Also including comparisons of before/after would be useful.

I added a simple timestamp around `encryptData()` and I'm not seeing any improvements
CleanShot 2024-09-17 at 05 14 08

### Single Answer Applet
|Run|Time Before (ms)|Time After (ms)|
|-|-|-|
|1 | 3216 | 3225 |
|2 | 3130 | 3116 |
|3 | 3115 | 3221 |
|4 | 3117 | 3117 |

### Applet with 4 answers
|Run|Time Before (ms)|Time After (ms)|
|-|-|-|
|1 | 3123 | 3227 |
|2 | 3121 | 3121 |


Variation between runs looks within margin of error. The improved performance appears to be due to removing the `wait(delay)`, which probably should also remove the calls to `this.notifyAsync` from `UploadProgressObservable` and instead use `this.notify`

felipeMetaLab:
I reverted to the old version of encryption.

ChaconC (approved):

PR #859, opened 2024-09-18
feat: display activity assignment target (M2-7407, M2-7408)

qiushihe:
### 📝 Description

🔗 [Jira Ticket M2-7407](https://mindlogger.atlassian.net/browse/M2-7407)
🔗 [Jira Ticket M2-7408](https://mindlogger.atlassian.net/browse/M2-7408)

This PR implements the display of activity assignment target, when the target is not the same as the respondent.

### 📸 Screenshots

**Activities Listing Screen**

iPhone 15 Pro Max 2024-09-18 15-16-08

### 🪤 Peer Testing

1. Sign into the admin app using `full-account-1`
2. Create an applet with 3 manual-assigned activities: `activity-1`, `activity-2` and `activity-3`
3. Invite `full-account-2` to the applet
4. Add 2 more limited accounts: `limited-1` and `limited-2`
5. Assign `activity-1` to `full-account-2` with `limited-1` as target
6. Assign `activity-2` to `full-account-2` with `limited-2` as target
7. Assign `activity-3` to `full-account-2` with `full-account-2` (i.e. "self") as target
8. Sign into the mobile app as `full-account-2`
9. Click on the applet from the applets listing screen
10. At this point, the assignment target badge should be visible for `activity-1` and `activity-2`, but not `activity-3`
11. Start either `activity-1` or `activity-2`
12. At this point, the assignment target badge should be visible on every step of the activity
13. Start `activity-3`
14. At this point, the steps of the activity should not have any assignment target badge

### ✏️ Notes

This PR also implements parts of: 🔗 [Jira Ticket M2-7408](https://mindlogger.atlassian.net/browse/M2-7408)

For [M2-7408](https://mindlogger.atlassian.net/browse/M2-7408), both the assignment target badge and the assignment target banner are implemented in this PR. However, since the mobile app is currently missing the "welcome" screen, the banner is temporarily placed in `ActivityItem.tsx`, and is hidden. When the related bug ticket is being worked on (https://mindlogger.atlassian.net/browse/M2-7917), the banner should be moved into the correct place on the welcome screen.

iPhone 15 Pro Max 2024-09-18 15-16-22

[M2-7408]: https://mindlogger.atlassian.net/browse/M2-7408?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


qiushihe:
This PR also implements parts of: 🔗 [Jira Ticket M2-7408](https://mindlogger.atlassian.net/browse/M2-7408)

For [M2-7408](https://mindlogger.atlassian.net/browse/M2-7408), both the assignment target badge and the assignment target banner are implemented in this PR. However, since the mobile app is currently missing the "welcome" screen, the banner is temporarily placed in `ActivityItem.tsx`, and is hidden. When the related bug ticket is being worked on (https://mindlogger.atlassian.net/browse/M2-7917), the banner should be moved into the correct place on the welcome screen.

farmerpaul (changes requested):
Hey @qiushihe, before I go too deep into your code, I thought I'd share some feedback in testing this PR as I think it may entail some (hopefully minimal) refactoring.

I have many activities, flows, and assignments locally, since I've been working on MI Assign for some time. In the Web App, each activity assignment is listed as a distinct item, including those for in-progress activities and flows.

However in the mobile app, I still only see **one instance** of each activity/flow, despite the fact many of these activities/flows have several assignments.

| Web | Mobile |
| - | - |
| CleanShot 2024-09-19 at 16 26 56@2x | CleanShot 2024-09-19 at 16 40 36@2x |

Being able support increased granularity of activites/flows, including in-progress activities/flows – now tracking the assignment as well – was a non-trivial part of the work I did for [M2-7405](https://mindlogger.atlassian.net/browse/M2-7405) and [M2-7406](https://mindlogger.atlassian.net/browse/M2-7406).

I see that this was not made explicit in the Jira ticket, and I apologize for not thinking to mention it. Maybe I thought that since you'd reviewed my work on M2-7405 you'd have clocked that, but I shouldn't have assumed!

Anyway I suspect adding that support will just require building upon your current work. Please let me know if you have any questions around this change!

mbanting:
> Hey @qiushihe, before I go too deep into your code, I thought I'd share some feedback in testing this PR as I think it may entail some (hopefully minimal) refactoring.
> 
> I have many activities, flows, and assignments locally, since I've been working on MI Assign for some time. In the Web App, each activity assignment is listed as a distinct item, including those for in-progress activities and flows.
> 
> However in the mobile app, I still only see **one instance** of each activity/flow, despite the fact many of these activities/flows have several assignments.
> 
> Web	Mobile
> CleanShot 2024-09-19 at 16 26 56@2x	CleanShot 2024-09-19 at 16 40 36@2x
> Being able support increased granularity of activites/flows, including in-progress activities/flows – now tracking the assignment as well – was a non-trivial part of the work I did for [M2-7405](https://mindlogger.atlassian.net/browse/M2-7405) and [M2-7406](https://mindlogger.atlassian.net/browse/M2-7406).
> 
> I see that this was not made explicit in the Jira ticket, and I apologize for not thinking to mention it. Maybe I thought that since you'd reviewed my work on [M2-7405](https://mindlogger.atlassian.net/browse/M2-7405) you'd have clocked that, but I shouldn't have assumed!
> 
> Anyway I suspect adding that support will just require building upon your current work. Please let me know if you have any questions around this change!

Oh no. Looking at the AC for [M2-7407](https://mindlogger.atlassian.net/browse/M2-7407) this was clearly not stated and is missing entirely. There's actually several differences in the AC between this ticket and the web version, [M2-7405](https://mindlogger.atlassian.net/browse/M2-7405) even if there shouldn't be. 

Left a [comment](https://mindlogger.atlassian.net/browse/M2-7407?focusedCommentId=24979) on the ticket for Cari. Let's take the conversation there and perhaps huddle in the morning.

qiushihe:
@farmerpaul I just pushed up all the refactor works I've been working on for the past several days. And I think I got everything covered. So this is now ready for review.

Please take a look when you get a chance. Thanks!

farmerpaul (changes requested):
Just noted a couple remaining issues, which we've discussed via Slack:
- This work isn't behind the feature flag `enableActivityAssign` yet – please make sure functionality is unchanged when this feature flag is turned off.
- When an activity is set to auto-assign = false, and it has no manual assignments, I was still able to see the activity in the list of activities to be able to take. Unassigned activities should in fact be omitted from the list.

I know you're working hard on this and I appreciate your persistence! ❤️

qiushihe:
Hi @farmerpaul I finally finished all the changes, including the feature flag and those other changes you requested.

The feature flag is only checked at 1 place: https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/859/files#diff-413f489f443094109b8ee79a6c84cf6f99200e2ad44d0869526cb4d652b0cb75R135 so you only have to edit that if you want to toggle to see the differences.

Please take another look. Thanks!

I'm also happy to do more walkthrough/huddle/whatnot too 😄

src/widgets/activity-group/model/services/ActivityGroupsBuildManager.ts
```diff
@@ -107,54 +123,90 @@ const createActivityGroupsBuildManager = (logger: ILogger) => {
...
+          );
+        }
 
-    entityEvents = sort(entityEvents);
+        return entityEvent;
+      })
+      .filter(x => x.event.scheduledAt)
+      .filter(x => !x.entity.isHidden);
+
+    const sortedEntityEvents = sort(entityEvents);
```
farmerpaul, line 188: It looks like we each refactored `ActivityGroupsBuildManager` slightly differently. I think most of the differences are in how we each chose to loop through the array of events, but there's also one important **functional difference** that I think needs to be fixed in the mobile app: in the web app, if an activity is both `autoAssign: true`, _and_ has manual assignments, then the web app takes both into consideration and includes a self-report for that activity, and any additional manual assignments. It appears that in the mobile app, however, if it's discovered that the activity has manual assignments, then the `autoAssign` setting is disregarded.

Though it's true that in the admin app it's hard to have an activity that's both auto-assigned and have manual assignments (because the manual assignments panel disables choosing activities set to `autoAssign: true`), it's **not impossible**: You can set an activity to `autoAssign: false`, add manual assignments, then set it to `autoAssign: true` again.

Kenroy and I discussed this a while back and agreed it was important to anticipate this possibility and support both assignment states, both because the BE allows it, and because it may indeed becomes desirable from a requirements perspective to explicitly support it after we've seen how assignments are used out in the wild. Then if we did want to better support it, then we'd only need to make changes in the admin app, not the respondent web & mobile apps as well.

Let me know if you agree with adding support for activities that are both auto-assigned, and have manual assignments – and if so, could you make the required adjustment? 🙏🏻 

This is a more minor point but still worth noting: when I approached refactoring this function, I observed how many times the array was being reiterated and reprocessed, in a way that was both less performant and harder to read, than simply iterating through it just once via a simple `for` loop. So I converted it to the latter. I think you've reduced the number of iterations, but there's still 2 – a `reduce` followed by a `map`. I still find my `for` loop easier to follow (while I have a good feel for `reduce` now, it still sometimes hurts my brain!).

Did you reference my changes to the web app's `ActivityGroupsBuildManager` when you refactored this function, or did you feel there wasn't enough in common between the two implementations for it to be worth it?

farmerpaul: For reference, here's the version of the [loop in the web app](https://github.com/ChildMindInstitute/mindlogger-web-refactor/blob/26d0036cd1d0bebcd92284134a136cab9862ecb3/src/widgets/ActivityGroups/model/services/ActivityGroupsBuildManager.ts#L104-L141):

```tsx
    const eventEntities: EventEntity[] = [];
    const calculator = EventModel.ScheduledDateCalculator;

    for (const event of events) {
      const entity = idToEntity[event.entityId];
      if (!entity || entity.isHidden) continue;

      event.scheduledAt = calculator.calculate(event);
      if (!event.scheduledAt) continue;

      if (params.assignments) {
        const assignments = idToAssignments[entity.id] ?? [];

        // Add auto-assignment if enabled for this activity/flow
        if (entity.autoAssign) {
          eventEntities.push({ entity, event, targetSubject: null });
        }

        // Add any additional assignments (without duplicating possible auto-assignment)
        for (const { respondentSubject, targetSubject } of assignments) {
          const isSelfAssign = respondentSubject.id === targetSubject.id;
          if (entity.autoAssign && isSelfAssign) continue;

          eventEntities.push({
            entity,
            event,
            targetSubject: isSelfAssign ? null : targetSubject,
          });
        }
      } else {
        // Assignments disabled
        eventEntities.push({
          entity,
          event,
          targetSubject: null,
        });
      }
    }
```

src/entities/activity/lib/types/activityAssignment.ts
```diff
@@ -0,0 +1,29 @@
...
+  userId: string | null;
+  secretUserId: string;
+  firstName: string;
+  lastName: string;
+  nickname: string;
+  tag: string;
+  lastSeen: string | null;
+};
+
+type AssginmentBase = {
```
farmerpaul, line 12:
```suggestion
type AssignmentBase = {
```

src/features/pass-survey/ui/ActivityItem.tsx
```diff
@@ -400,6 +448,16 @@ function ActivityItem({
...
       >
+        {assignment && assignment.respondent.id !== assignment.target.id && (
+          <XStack mx={16} mb={12}>
+            <ActivityAssignmentBadge
+              assignment={assignment}
+              accessibilityLabel="item_display_assignment"
+            />
+            <Box flexGrow={1} flexShrink={1} />
+          </XStack>
+        )}
```
farmerpaul, line 457: The multi-informant badge is meant to be inserted into the item's markdown, below any paragraphs that just contain media (images, etc.) and directly above the first paragraph that contains text. Check out the web app to see how this was done; hopefully you can just reuse the same utility function [`insertAfterMedia`](https://github.com/ChildMindInstitute/mindlogger-web-refactor/pull/534/files#diff-b71cf5aa4da256f58f6fca6f8557c99cba908811493da204113a2a90445d1111) from the web app to achieve this! You can even copy the unit tests over.

src/entities/activity/model/mappers.input.mock.ts
```diff
@@ -10,6 +10,7 @@ const basicAppletDetails = {
   isSkippable: false,
   isReviewable: false,
   isHidden: false,
+  autoAssign: false,
```
farmerpaul, line 13: When assignments feature flag is enabled in production, all existing activities and flows will have `autoAssign: true` by default. Also, new activities and flows will have `autoAssign: true`. In the web apps where I've had to add the `autoAssign` property to mock activity data, I've defaulted that property value to `true` because it's the default value as well as the more common use case to test (until people actually start using manual assignments). Wondering what your reason is for choosing `false` here and in all other mocks by default. Perhaps it has no actual impact in the end since the existing unit tests aren't doing anything with that value.

package.json
```diff
@@ -164,7 +163,6 @@
     "eslint-plugin-react": "^7.33.2",
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-unused-imports": "^3.0.0",
-    "husky": "^8.0.3",
```
farmerpaul, line 167: 😍

src/features/auth/lib/clearEntityRecordStorages.ts
```diff
@@ -1,5 +1,5 @@
-import { getDefaultMediaFilesCleaner } from '@app/entities/activity/lib/services/mediaFilesCleanerInstance';
-import { getDefaultStorageInstanceManager } from '@app/shared/lib/storages/storageInstanceManagerInstance';
+import { getDefaultMediaFilesCleaner } from '@entities/activity/lib/services/mediaFilesCleanerInstance.ts';
+import { getDefaultStorageInstanceManager } from '@shared/lib/storages/storageInstanceManagerInstance.ts';
```
farmerpaul, line 2: Is there a reason to add the `.ts` extension to these import paths?

Same thing [here](https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/859/commits/4af9ecb54895c8e6b51e12b5008a584a06359ec6#diff-671e9b4b9ec2d18a12b3dd87d343a500e580f98c83de6fa62cbf79e8942f2757R1), [here](https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/859/commits/4af9ecb54895c8e6b51e12b5008a584a06359ec6#diff-a83f923d379c68ecb76a5b0f9307a8ab4b3971a6c8e2e15f8b56f95121ba0809R1-R3), [here](https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/859/commits/4af9ecb54895c8e6b51e12b5008a584a06359ec6#diff-2b00f31120d77882c5b15cd11083e8e063f5e3339dc813d7674034e69fa3fb24R31), and [here](https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/859/commits/4af9ecb54895c8e6b51e12b5008a584a06359ec6#diff-46a7e9d5ae997ecd9b3d9fe46b52910b3fbc1f65ffe8e3f5e222ab488faaef10R19).

farmerpaul (changes requested):
I have a bit more feedback @qiushihe. I reviewed commit by commit to help me digest what the changes were more easily (and prevent my browser from to slowing to crawl during review). But even so, I will admit I only skimmed some of the more bulky commits in order to maintain my sanity, as it's a lot for me to absorb!

I really appreciate the walkthroughs you've given both via huddle and the demo, and I understand the majority of the refactoring changes you've made. But nevertheless I may have missed some details – hopefully QA will catch anything I've missed.

.eslintrc.js
```diff
@@ -127,7 +127,7 @@ module.exports = {
         '@typescript-eslint/unbound-method': 'warn',
         '@typescript-eslint/restrict-plus-operands': 'warn',
         '@typescript-eslint/no-unsafe-return': 'warn',
-        '@typescript-eslint/restrict-template-expressions': 'warn',
+        '@typescript-eslint/restrict-template-expressions': ['warn', { allowNever: true }],
```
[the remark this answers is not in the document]
qiushihe: Yeah I actually okay with just removing this rule as well.

src/features/applets-refresh/ui/AppletsRefresh.tsx
```diff
@@ -29,7 +29,7 @@ const AppletsRefresh: FC<Props> = props => {
       completions,
       LogTrigger.PullToRefresh,
     );
-    Logger.send();
+    Logger.send().catch(err => Logger.error(err as never));
```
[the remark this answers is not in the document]
qiushihe: Sorry this was due to my initial mis-understanding of the code. It should be `.catch(console.error)` here. So the code would try to use `Logger` to send, and if that fails, log the error with `console.error`.

src/entities/activity/model/mappers.input.mock.ts
qiushihe: I was mostly trying to prevent too many tests from failing. But that was just a preemptive measure. I feel it should be okay to change this to `true`. I'll give it a try.

src/features/auth/lib/clearEntityRecordStorages.ts
qiushihe: I'm actually not sure. Both VScode and WebStorm did that in separate instances. I'm not sure why that's the case. But I can manually remove those extensions from important statements (it should be easy to find them all).

package.json
qiushihe: I believe you'd still need to remove the husky stuff locally (if you haven't already done so)?

src/features/auth/lib/clearEntityRecordStorages.ts
qiushihe: Actually we're using `eslint-plugin-import`, so we can just add a `"import/extensions": "never"` rule 💁‍♀️ 
 https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/extensions.md

qiushihe: Turns out that rules doesn't support auto-fix 😵 I guess I'll just manually remove them.

src/widgets/activity-group/model/services/ActivityGroupsBuildManager.ts
qiushihe: Honestly I did not reference the changes in the web app for this part of the changes, mostly because the way I did all the changes were very incremental in order to not break anything. So I didn't realize there are some functional differences between the 2 versions. But I'm happy to update my code to do the same things.

With regard to the reduce followed by map, that's also just due to the incremental nature of the changes. The map was already there originally, which I didn't originally wanted to touch. But throughout the refactor process, the stuff in the reduce got added there and things just sort of settled down to the way they are now. Plus, originally I didn't have any working tests to back me up, so I was just being extra careful to not break anything that's existing. But I'm happy to combine the 2 to make them more efficient.

qiushihe: Actually for the first part about functional parity, it's completely my bad for not catching it, regardless if I referenced your code or not.

The reasons I said that is because my code is basically:

```
if (entityAssignments.length <= 0) {
  if (entity.autoAssign) {
    // action block A1
  } else {
    // action block A2
  }
} else {
  // action block B
}
```

... and had I been more careful, I would have noticed the outer `else`'s action block is missing the same `autoAssign` check.

My normal approach to coding this sort of nested conditions is to just fan out all the branches, then fill in the details:

```
if (entityAssignments.length <= 0) {
  if (entity.autoAssign) {
    ...
  } else {
    ...
  }
} else {
  if (entity.autoAssign) {
    ...
  } else {
    ...
  }
}
```

This way it's easier for me to catch any "holes" in the logical branching. But I didn't do that this time, and it immediately came back the bit me in the butt.

So anyway, I'll get this fixed asap.

src/features/pass-survey/ui/ActivityItem.tsx
qiushihe: It seems to work pretty well 💁‍♀️ 

iPhone 15 Pro Max 2024-10-06 12-02-12

qiushihe:
@farmerpaul okay I think I got everything covered. Please take another look. Thanks!

farmerpaul (approved):
Hey Billie, great work on all of this!! Thank you for making those changes. I have one last request, but preapproving because I have total faith you'll be able make this change without further approval, and I'd like to speed along getting this into QA.

Another change I made in the web app was to sort the activities/flows according to the AC. The AC says to sort only by subject first name, but I opted to sort both by first name, then last initial.

The original sort method is also inefficient, performing two calls to `.filter()`, then two calls to `.sort()`. I optimized it to use a single `.sort()` call, as well as taking into account target subject. Here's my version, feel free to copy-paste/adapt:

```tsx
  const sort = (eventEntities: EventEntity[]) => {
    eventEntities.sort((a: EventEntity, b: EventEntity) => {
      const aIsFlow = a.entity.pipelineType === ActivityPipelineType.Flow;
      const bIsFlow = b.entity.pipelineType === ActivityPipelineType.Flow;

      // Order flows first
      if (aIsFlow && !bIsFlow) {
        return -1;
      } else if (!aIsFlow && bIsFlow) {
        return 1;
      }

      // Then order by entity order
      const orderDiff = a.entity.order - b.entity.order;
      if (orderDiff !== 0) return orderDiff;

      // Then order self-reports first
      if (!a.targetSubject) {
        return -1;
      } else if (!b.targetSubject) {
        return 1;
      }

      // Then order by target subject first name
      const firstNameDiff = a.targetSubject.firstName.localeCompare(b.targetSubject.firstName);
      if (firstNameDiff !== 0) return firstNameDiff;

      // Then order by target subject last name
      return a.targetSubject.lastName.localeCompare(b.targetSubject.lastName);
    });
  };
```

Then it's just called using:

```tsx
    sort(eventEntities);
```

PR #862, opened 2024-10-02
chore: changing the crypto library (M2-7904)

felipeMetaLab:
### 📝 Description
🔗 [Jira Ticket M2-7904](https://mindlogger.atlassian.net/browse/M2-7904)

Adding new library to encryption and decryption.
The crypto lib that the app was using was hard coded, so I removed the hard coded library and changed for react-native-crypto.
react-native-crypto was the best solution because it already has support for all the methods that the application was already using, that is, this way we don't have to make big changes on code, only a proper configuration. 

### 📸 Screenshots

https://github.com/user-attachments/assets/891b5f32-4775-4d4b-96ba-e203fca3e220

### 🪤 Peer Testing
1 - launch the app, backend and admin
2 - create an applet and one activity
3 - access the activity via App  and go through all steps
4 - check if the app was able to render the pages and encrypt the answer.
5 - check if the answers comes to admin and admin is able to decrypt it.
6 - check if backend is receiving the encrypted answers and saving successfully
7 - check if you are able to make login successfully

### ✏️ Notes
I tried different libraries such as crypto-js, crypto, and a bunch of others, the one that fits best is react-native-crypto, its has support for all methods the app is using already.


felipeMetaLab:
I am still fixing the tests and the lint.

PR #867, opened 2024-10-10
feat: changing the crypto library (M2-7904)

felipeMetaLab:
### 📝 Description
🔗 [Jira Ticket M2-7904](https://mindlogger.atlassian.net/browse/M2-7904)

Adding new library to encryption and decryption.
The crypto lib that the app was using was hard coded, so I removed the hard coded library and changed for crypto-browserify.

This version handles the installation of crypto-browserify directly using yarn add crypto-browserify. To ensure it works properly with React Native, I needed to make some configurations. For example, I set up streams through a polyfill and defined two global variables: Buffer and process.

### 📸 Screenshots

https://github.com/user-attachments/assets/891b5f32-4775-4d4b-96ba-e203fca3e220

Uploading android test.mov…

### 🪤 Peer Testing
1 - launch the app, backend and admin
2 - create an applet and one activity
3 - access the activity via App  and go through all steps
4 - check if the app was able to render the pages and encrypt the answer.
5 - check if the answers comes to admin and admin is able to decrypt it.
6 - check if backend is receiving the encrypted answers and saving successfully
7 - check if you are able to make login successfully

### ✏️ Notes
I tried different libraries such as crypto-js, crypto, and a bunch of others, the one that fits best is react-native-crypto, its has support for all methods the app is using already.


src/shared/lib/encryption/crypto.ts
ChaconC, line 1: You can remove this unnecessary file and import crypto-browserify directly from EncryptionManager

src/shared/lib/encryption/EncryptionManager.test.ts
```diff
@@ -135,7 +138,9 @@ describe('Encryption', () => {
 
       const generatedPrivateKey = encryptionManager.getPrivateKey(userParams);
 
-      expect(generatedPrivateKey).toStrictEqual(result);
+      const expectedPrivateKey = encryptionManager.getPrivateKey(userParams);
```
ChaconC, line 141: The way you updated this test is incorrect now.
`expectedPrivateKey` and `generatedPrivateKey` are both the result from the same call to `encryptionManager.getPrivateKey`

globals.d.ts
```diff
@@ -3,3 +3,5 @@ type TimeoutId = ReturnType<typeof setTimeout>;
 type IntervalId = ReturnType<typeof setInterval>;
 
 type Maybe<TValue> = TValue | null | undefined;
+
+declare module 'crypto-browserify';
```
ChaconC, line 7: Is this global module required?

react-native.config.js
```diff
@@ -1,4 +1,6 @@
+const extraNodeModules = require('node-libs-browser');
```
ChaconC, line 1: I think this was from a previous implementation test, is it still required with this implementation?

ChaconC (approved):

PR #869, opened 2024-10-16
chore: auto version bump commit

ChaconC:
- [ ] OSS packages added to MindLogger [open source credit page](https://mindlogger.atlassian.net/jira/servicedesk/projects/MLA/knowledge/articles/340623543?spaceKey=MLA)

### 📝 Description

Automatically adds `chore: bump version X.X.X (xxxx)` commit when doing yarn bump

### 📸 Screenshots

CleanShot 2024-10-16 at 14 00 50

### 🪤 Peer Testing

- Run `yarn bump`
- Check last commit message

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -30,9 +30,23 @@ package = load_json(json_path: "./package.json")
   )
 end
 
+desc 'Automatically add commit message including build number.'
+private_lane :commit_version do |options|
+  package = load_json(json_path: "./package.json")
+  xcodeproj_path = 'ios/MindloggerMobile.xcodeproj'
+  version_number = package['version']
+  build_number = get_build_number(xcodeproj: xcodeproj_path)
+
+  git_commit(path: [
+    './android/app/build.gradle',
+    './ios/MindloggerMobile.xcodeproj/project.pbxproj'
+    ], message: "chore: bump version #{version_number} (#{build_number})"
+  )
+end
 
 desc 'Bump build numbers, and set the version to match the pacakage.json version.'
 lane :bump_version do |options|
   increment_version_android
   increment_version_ios
+  commit_version
 end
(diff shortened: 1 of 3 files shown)


fastlane/Fastfile
```diff
@@ -30,9 +30,23 @@ package = load_json(json_path: "./package.json")
...
+  build_number = get_build_number(xcodeproj: xcodeproj_path)
+
+  git_commit(path: [
+    './android/app/build.gradle',
+    './ios/MindloggerMobile.xcodeproj/project.pbxproj'
+    ], message: "chore: bump version #{version_number} (#{build_number})"
+  )
+end
 
 desc 'Bump build numbers, and set the version to match the pacakage.json version.'
```
mbanting, line 47: Small typo
```suggestion
desc 'Bump build numbers, and set the version to match the package.json version.'
```

mbanting (commented):
Can we make the commit optional? 
Or we can have this by default but add a parameter to bypass it

There have been a couple of occasions where main/release wasn't merged back to dev, so dev was a couple of build numbers behind. That required having to run
`yarn bump` multiple times to ensure it's higher than the most recent version.

With this, each time `yarn bump` is called, a commit would be added. Ideally, we can bypass the commit until the last final one.

ChaconC:
Sure thing @mbanting, I'll make the change to make it optional

fastlane/Fastfile
ChaconC: Done! you can now pass `skipCommit:true` and `increment:x` like `yarn bump skipCommit:true increment:5`
Default behavior is to increment by 1 and generate the commit message

ChaconC:
@mbanting 
Updated! you can now pass skipCommit:true and increment:x like yarn bump skipCommit:true increment:5
Default behavior is to increment by 1 and generate the commit message

PR #870, opened 2024-10-21
Fix/m2 7967 fix input tap detection

felipeMetaLab:
### 📝 Description

🔗 [Jira Ticket M2-7967](https://mindlogger.atlassian.net/browse/M2-7967)
#### Solution Explanation:
DispatchQueue:
Main Thread for UI Updates, In iOS development, all UI updates must occur on the main thread. UIKit is not thread-safe, and updating UI elements from a background thread can lead to unpredictable behavior or crashes and  Wrapping UI updates in DispatchQueue.main.async ensures that the enclosed code executes on the main thread, regardless of which thread the enclosing method was called from.
 what do we gain? By explicitly dispatching UI updates to the main thread, we eliminate the risk of UI related issues due to threading.

Timer Instead of Timer.scheduledTimer:
Using Timer Instead of Timer.scheduledTimer Timer.scheduledTimer creates a timer and automatically schedules it on the current run loop in the default mode.
If the code calling Timer.scheduledTimer is not on the main thread, the timer may be scheduled on a background run loop, leading to timing issues or the timer not firing at all and Timers scheduled in the default run loop mode may not fire during certain user interactions because the run loop switches modes during these interactions.  Timer Initialization Without Scheduling:
* Manual Scheduling: By using Timer(timeInterval: ...), its create a timer instance without automatically scheduling it on a run loop.
* Control Over Run Loop and Mode: We can then manually add the timer to the desired run loop and specify the run loop mode, giving us better control over its behavior.

  Using RunLoop.main.add:
After creating a Timer instance without scheduling it, I need to add it to a run loop to start it. 
* .default: The default mode for the main run loop. Timers in this mode may not fire during user interactions.
* .common: A set of modes that includes the default mode and modes used during user interactions. Timers added in this mode continue to fire even when the run loop switches modes due to user interactions.

#### Simple Choice Reaction Time task:

https://github.com/user-attachments/assets/7b9db765-eabf-4982-91bc-e022270a618d

### 🪤 Peer Testing
* lauch the app
* Run task on an iOS device
Before:
Input will be ignored the longer you wait to tap the button (disabled state)
Expected:
now you will be able to click all the times.

### DIFF
--- a/ios/FlankerNativeComponents/GameManager.swift
+++ b/ios/FlankerNativeComponents/GameManager.swift
@@ -308,7 +308,7 @@ class GameManager {
     if !isFirst {
       countTest += 1
     }
-    
+
 
     if gameParameters.showFixation {
       if let image = URL(string: gameParameters.fixation), gameParameters.fixation.contains("https") {
@@ -406,7 +406,6 @@ private extension GameManager {
 
   func isEndGame() -> Bool {
     guard let gameParameters = gameParameters else { return false }
-
     if countTest == gameParameters.trials.count {
       let sumArray = arrayTimes.reduce(0, +)
       var avrgArray: Int = 0
--- a/ios/FlankerNativeComponents/GameManager.swift
+++ b/ios/FlankerNativeComponents/GameManager.swift
@@ -192,9 +192,7 @@ class GameManager {
   func checkedAnswer(button: SelectedButton) {
     invalidateTimers()
 
-    DispatchQueue.main.async {
-      self.delegate?.setEnableButton(isEnable: false)
-    }
+    delegate?.setEnableButton(isEnable: false)
 
     guard let gameParameters = gameParameters else { return }
     guard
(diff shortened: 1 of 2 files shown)


ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -191,7 +191,11 @@ class GameManager {
 
   func checkedAnswer(button: SelectedButton) {
     invalidateTimers()
-    delegate?.setEnableButton(isEnable: false)
+
+    DispatchQueue.main.async {
+      self.delegate?.setEnableButton(isEnable: false)
```
ChaconC, line 196: Since this is a delegate function you can wrap the `setEnableButton` function in FlankerView.swift instead of here to avoid repeating this multiple times.

Same goes for `updateText`.

ChaconC (changes requested):
Thanks for this @felipeMetaLab, I tested a 15 point SRT task and was able to get all button taps recorded correctly, both by tapping really fast and by waiting 1second (average).

|Fast|Slow|
|-|-|
|IMG_0DEA7413ABE7-1|IMG_D116625404AD-1

I left a code change suggestion but this looks promising.

ChaconC (approved):

PR #873, opened 2024-10-22
feat: Prevent starting of unsupported activities; redirect to web app when possible (M2-7925, M2-8059)

JustinNusca:
### 📝 Description

🔗 [M2-7925](https://mindlogger.atlassian.net/browse/M2-7925): [FE][Phrase Builder] Prevent mobile app crash when selecting Web Only Item types
🔗 [M2-8059](https://mindlogger.atlassian.net/browse/M2-8059): [FE][Phrase Builder] Add “Please complete on the MindLogger web app” for Phrase Builder activities in the mobile app

This PR addresses an issue where an activity or flow that contains items that are unsupported on the app may be started, which caused the app to crash.

I handled this by updating the `InProgressActivityScreen` to check for and validate the `responseType` values contained within an activity or flow, and redirect back to the relevant applet details page if it contained an unsupported `responseType`.

The list of activities and flows has also been updated. Now, if an activity is unsupported on the app, but supported on web, a Chip alerting the user appears on the list item. When pressed, the user is redirected to the Web App's relevant applet details page in their browser.

If an activity or flow contains `responseTypes` that are altogether unsupported, the list item is disabled entirely.

To implement these changes, I:

- [Added a `useBaseInfo` query and hook.](https://github.com/ChildMindInstitute/mindlogger-app-refactor/commit/ad08240799198a3412dd48337e3decf944977e10) 
  - This fetches from the `/applets/{appletId}/base_info` endpoint, which contains `containsResponseType` fields for activities.
  - The hook additionally transforms the returned data to include a `responseTypes` object, which includes key-value pairs of activity/flow IDs and their child item responseTypes.
- [Added `lib/constants/responseTypes` & `lib/utils/responseTypes` files.](https://github.com/ChildMindInstitute/mindlogger-app-refactor/commit/eddc8cdcce4a0804c58e9b60049bbeb7754755b0) 
  - The backend does not authoritatively declare which clients support which `responseTypes` so these constants and their corresponding utility functions are functionally equivalent to the constants defined in the admin app, [here](https://github.com/ChildMindInstitute/mindlogger-admin/blob/01b2ba07a72cf04cdb4f557eb8bf79f3d512ec8b/src/shared/utils/responseType.ts).
  - This can be used to determine if an activity may be completed on a given client.
- [Add additional ResponseType type values.](https://github.com/ChildMindInstitute/mindlogger-app-refactor/commit/728bbeb7c2fe768be2ded3402cbc44ad4fe6ccaa) 
  - This adds `phrasalTemplate` as a possible responseType value.
  - I also sorted the list alphabetically.
- [Update `InProgressActivityScreen` to validate responseTypes.](https://github.com/ChildMindInstitute/mindlogger-app-refactor/commit/af03d2b87c177d9765d84bf40ace67c39e80dc3d) 
  - Use `useBaseInfo` to fetch responseTypes, and validate support for the current activity or flow.
  - If the responseType is unsupported, redirect back to the applet details page.
- [Add new shared UI components: Chip and ExclamationIcon.](https://github.com/ChildMindInstitute/mindlogger-app-refactor/commit/266ef02211fcb5aba8d323e5d42587d77dcaeec6)
  - These were required to implement the new `Chip` UI on Activity list items.
  - I configured the `Chip` component to support a variant prop that may match the styles used in the web app (the admin app uses a different theme), but I did not add all known variants, since they're mostly unused for now. Support is there to easily add it if/when we need it.
- [Update behaviour when activities/flows contain unsupported items.](https://github.com/ChildMindInstitute/mindlogger-app-refactor/commit/57c6dbaa87785b69b0b90042c85e4781323acd8d) 
  - When an activity or flow is only supported on the web, pressing on the item will open the browser and direct the user to the given applet details page.
  - When an activity or flow is only supported on the web, a Chip will notify the user that the item must be completed on the web app.
  - When an activity or flow is otherwise unsupported, the item is disabled entirely.

### 📸 Screenshots

| Before | After |
|-|-|
| Simulator Screenshot - iPhone 16 Pro - 2024-10-22 at 14 00 21 | Simulator Screenshot - iPhone 16 Pro - 2024-10-22 at 14 00 08 |
| before | after |

### 🪤 Peer Testing

For an applet that contains both an Activity, and an Activity Flow, with "Phrase Builder" item types…

1. From the `ActivityListScreen`, view the list of activities and activity flows.
2. Observe that list items that contain the phrase builder are marked with a Chip that indicates the activity or flow must be completed on the web app.
3. Press a "web only" activity or flow. Observe that you are redirected to the browser to complete the activity or flow.
4. Return to the app. Observe that you are still on the `ActivityListScreen` and can continue to use the app.

[M2-7925]: https://mindlogger.atlassian.net/browse/M2-7925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[M2-8059]: https://mindlogger.atlassian.net/browse/M2-8059?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/entities/activity/ui/ActivityCard.tsx
```diff
@@ -111,6 +111,12 @@ export const ActivityCard: FC<Props> = ({ activity, disabled, onPress }) => {
             </Text>
           )}
 
+          {isWebOnly && (
+            <Chip icon={ExclamationIcon} variant="warning">
+              Please complete on the MindLogger web app
```
farmerpaul, line 116: This should be internationalized via `useTranslation`.

src/shared/lib/utils/responseTypes/responstTypes.test.ts
farmerpaul, line 1: Nit: can this be renamed to `responseTypes.test.ts`?

src/shared/ui/Chip.tsx
```diff
@@ -0,0 +1,61 @@
...
+};
+
+const StyledChip = styled(View, {
+  alignItems: 'center',
+  borderRadius: 4,
+  flexDirection: 'row',
+  gap: 4,
+  name: 'Chip',
+  px: 8,
+  py: 4,
```
farmerpaul, line 28: Nice Chip! Did you base this on an existing Figma component? If so, I'm happy to defer to that, but the only thing I could find was the one from the ML1 [Design System](https://www.figma.com/design/LXl9BAJSwtgWuzkxqOqvX8/ML1-Design-System?node-id=58697-179569&m=dev). There's a few very subtle differences, but I thought I'd offer my suggestions here (but point me to the right direction if there's another mobile design you referenced).

| Before | After |
| - | - |
| CleanShot 2024-10-23 at 20 03 10@2x | CleanShot 2024-10-23 at 20 03 40@2x |

Here, I changed:

```tsx
  borderRadius: 8,
  gap: 8,
  pl: 8,
  pr: 12,
  py: 6,
```

And then below, in the `Chip` component definition, I:
- Adapted `lineHeight` to be proportional to the smaller font size of `12px` (in Figma, line height is 20px but for a 14px font size, i.e. line height is 142% of font size)
- Font weight should be 400
- I added `flex={1}` to prevent wrapping text from sometimes overflowing the bounds of the chip (try on iPhone 15 or iPhone 15 Pro)

```tsx
      <Text fontSize={12} fontWeight="400" lineHeight={17} flex={1}>
```

farmerpaul (changes requested):
Hey @JustinNusca, amazing work on this. 💪🏻  I had a few small comments, see what you think.

src/shared/ui/Chip.tsx
JustinNusca: The styles here (padding, borderRadius, colors, etc) were a direct copy of the existing label in the web app — the linked task was to adapt the web UI to mobile.

_That_ label was itself styled as a one-off component, but I couldn't help but notice that it was identical in all respects except for colors to a number of variations throughout the web app. The ones I spotted were also built as one-offs.

I couldn't find anything in Figma or documentation elsewhere that matched the appearance of this pattern in the web app, but it's the pattern as it appeared _there_ that I attempted to formalize _here_. 

I saw the `Chip` you linked to, but it seemed distinct enough (different styles, as you point out), and had a lot more states that weren't reflected in the web-app. Maybe I should have named it something different, but I wasn't trying to replicate _that_ `Chip`, specifically 😅.

I went ahead and updated the styles to mostly match your suggestions, regardless — I think the after is much improved. I kept the horizontal padding equal, though, mostly to avoid inconsistent values if an icon were omitted.

src/shared/lib/utils/responseTypes/index.ts
```diff
@@ -0,0 +1,21 @@
+import { ResponseType } from '@app/shared/api/services/ActivityItemDto';
+
+import {
+  appSupportedResponseTypes,
+  universalSupportedResponseTypes,
+  webSupportedResponseTypes,
+} from '../../constants/responseTypes';
+
+export const getIsAppOnly = (responseType: ResponseType) =>
```
farmerpaul, line 9: Nit: I know this repo is called `mindlogger-app-refactor`, but I always thought that was a misnomer because all three FE projects are indeed apps – this one's a mobile app, while the other two are web apps. Would you be all right with using the language "mobile" here instead? I think we're already using that language in the Admin App.

```suggestion
export const getIsMobileOnly = (responseType: ResponseType) =>
```

src/shared/lib/utils/responseTypes/index.ts
```diff
@@ -0,0 +1,21 @@
...
+  appSupportedResponseTypes.includes(responseType);
+
+export const getIsWebOnly = (responseType: ResponseType) =>
+  webSupportedResponseTypes.includes(responseType);
+
+export const getSupportsApp = (responseType: ResponseType) =>
```
farmerpaul, line 15: Same here:
```suggestion
export const getSupportsMobile = (responseType: ResponseType) =>
```

farmerpaul (approved):
Thanks for updating the chip styling, and for explaining the basis for it – makes total sense to base it on the web app.

Just had one last nit about naming, but otherwise you got my approval! Great work as always! 💖

src/shared/lib/utils/responseTypes/index.ts
JustinNusca: Updated [here](https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/873/commits/acb8c088b64e02c11ece0aea244753383474a503)!

src/shared/lib/utils/responseTypes/index.ts
JustinNusca: Updated [here](https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/873/commits/acb8c088b64e02c11ece0aea244753383474a503)!

PR #875, opened 2024-10-24
feat: Omit Report Scores with Scoring Type "Score" from the Report Summary Screen (M2-8097)

sultanofcardio:

### 📝 Description

🔗 [Jira Ticket M2-8097](https://mindlogger.atlassian.net/browse/M2-8097)

This PR updates the Report Summary screen to omit report scores configured with scoring type `score`. These are scores that are linked to a subscale and are currently not supported as part of the report summary screen.

### 📸 Screenshots

N/A

### 🪤 Peer Testing

> [!NOTE]
> This PR depends on https://github.com/ChildMindInstitute/mindlogger-admin/pull/1935 in order to configure the activity from the admin app

In the admin app
1. Create an activity with a subscale containing a lookup table and an activity item (in its "Elements within Subscale" list)
2. Create a report score and set the score type to "Score"
3. Select the subscale from the dropdown list
4. Create another report score and set the score type to "Raw Score"
5. Save the activity

In the mobile app
1. Complete the activity you created above
2. Confirm that the report summary screen does not include the report score with score type set to "Score"

### ✏️ Notes

N/A


mbanting:
Tested on app and it seems to be working, although I do find it misleading that on the Scoring screen admins can choose Show Score Summary and Score type and the system ignores it. I'll post a comment on the ticket and followup with Cari.

.node-version
```diff
@@ -1 +1 @@
-18
+20.10.0
```
mbanting, line 1: was this intended to be updated?

mbanting (approved):
LGTM, just left a question. Also posted a product-related question but will follow that up in JIRA (any outcome of that would be a separate ticket & PR).

.node-version
sultanofcardio: It's intentional, but I realise that I forgot to update the README. I'll add that

PR #876, opened 2024-10-24
M2-7967: fix input tap detection

felipeMetaLab:
### 📝 Description

🔗 [Jira Ticket M2-7967](https://mindlogger.atlassian.net/browse/M2-7967)
#### Solution Explanation:
this is a continuation of https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/870. 
This time I am handling the enable and disable button.

#### Simple Choice Reaction Time task:

https://github.com/user-attachments/assets/7b9db765-eabf-4982-91bc-e022270a618d

### 🪤 Peer Testing
* lauch the app
* Run task on an iOS device
Before:
Input will be ignored the longer you wait to tap the button (disabled state)
Expected:
now you will be able to click all the times.


ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -190,16 +193,38 @@ class GameManager {
...
+      print("[GameManager.swift] User has already responded in this trial.") // <---- if is a better tracer handling let me know.
+      return
+    }
+    hasResponded = true  // Mark that the user has responded
 
-    delegate?.setEnableButton(isEnable: false)
+    invalidateTimers()
 
     guard let gameParameters = gameParameters else { return }
-    guard
```
felipeMetaLab, line 198: explanation Why I Removed It:
The guard statement caused the checkedAnswe method to return early if any of the variables were nil.
This led to situations where clicking the button didn't register the event because respondTouchButton or endTrialTimestamp hadn't been set yet

felipeMetaLab:
In this PR I left a lot of comment in the code to facilitate the test, but it will be removed once it get approved.

felipeMetaLab:
I left a lot of comment to facilitate the understanding, once it get approved I will remove all comments.
Any question let me know.

felipeMetaLab:
comments removed

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -333,27 +369,33 @@ class GameManager {
       delegate?.updateText(text: text, color: .black, font: Constants.bigFont, isStart: true, typeTime: .trial)
     }
 
+    setEndTimeViewingImage(time: CACurrentMediaTime(), isStart: true, type: .trial)
```
ChaconC, line 372: `setText` is also fired when fixations are displayed, can that lead to initializing incorrect values for `startTrialTimestamp` when it gets called from a different place?

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -190,16 +191,36 @@ class GameManager {
...
+      return
+    }
+
+    guard let startTrialTimestamp = startTrialTimestamp else {
+      print("[GameManager.swift]: startTrialTimestamp is nil in checkedAnswer") 
+      return
+    }
+
+    if respondTouchButton == nil {
+      respondTouchButton = CACurrentMediaTime()
```
ChaconC, line 215: Why are you changing the value set by `setEndTimeViewingImage`?

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -190,16 +191,36 @@ class GameManager {
...
+    }
+
+    guard let respondTouchButton = respondTouchButton else { return }
+
+    if endTrialTimestamp == nil {
+      endTrialTimestamp = respondTouchButton
```
ChaconC, line 221: Same here, you're changing the value set by a different function

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -333,27 +369,33 @@ class GameManager {
...
 
     guard
-      let startTrialTimestamp = startTrialTimestamp,
-      let endTrialTimestamp = endTrialTimestamp
+      let startTrialTimestamp = startTrialTimestamp
     else { return }
 
+    if endTrialTimestamp == nil {
+      endTrialTimestamp = CACurrentMediaTime()
+    }
```
ChaconC, line 393: What behavior is changing where previously we always had a end timestamp (from the `guard let`) that we now may or may not have a timestamp.

ios/FlankerNativeComponents/GameManager.swift
ChaconC: I think this is close to the cause but you're approach is backwards.

I understand the early return was problematic, but you should investigate the reason for nil values instead of arbitrarily adding values in here. The timing should be coming from elsewhere.

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -184,271 +124,257 @@ class GameManager {
...
-    let randSymbol = arc4random_uniform(UInt32(symbolArray.count))
-    let randLetters = arc4random_uniform(UInt32(lettersArray.count))
+    let sumArray = arrayTimes.reduce(0, +)
+    let avrgArray = arrayTimes.count > 0 ? sumArray / arrayTimes.count : 0
+    let procentsCorrect = Float(correctAnswers) / Float(countAllGame) * 100
 
-    return symbolArray[Int(randSymbol)] + String(lettersArray[Int(randLetters)]) + symbolArray[Int(randSymbol)]
+    clearData()
+
+    delegate?.updateText(text: "Game Over", color: .black, font: Constants.bigFont, isStart: false, typeTime: .feedback)
```
ChaconC, line 308: What is the purpose of this game over screen?

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -184,271 +124,257 @@ class GameManager {
...
+        break
       }
     }
   }
 
   func checkedAnswer(button: SelectedButton) {
+    guard !hasRespondedInCurrentTrial else { return }
+    hasRespondedInCurrentTrial = true
+    respondTouchButton = CACurrentMediaTime()
+    setEndTimeViewingImage(time: respondTouchButton!, isStart: false, type: .response)
```
ChaconC, line 136: This call to `setEndTimeViewingImage` is not required, right? Right now it's only validating to `break` from the change in line 127

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -184,271 +124,257 @@ class GameManager {
...
+          handleEndOfGame()
+          return
+      }
 
-    delegate?.setEnableButton(isEnable: true)
+      setEndTimeViewingImage(time: CACurrentMediaTime(), isStart: false, type: .fixations)
 
-    text = gameParameters.trials[countTest].stimulus.en
+      startTrialTimestamp = CACurrentMediaTime()
+      setEndTimeViewingImage(time: startTrialTimestamp!, isStart: true, type: .trial)
```
ChaconC, line 238: Same here, this call is not required since you changed it in line 112 to simply `break`

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -184,271 +124,257 @@ class GameManager {
...
   @objc func timeResponseFailed() {
     guard let gameParameters = gameParameters else { return }
 
     delegate?.setEnableButton(isEnable: false)
+
+    endTrialTimestamp = CACurrentMediaTime()
+    setEndTimeViewingImage(time: endTrialTimestamp!, isStart: false, type: .trial)
+
+    startFeedbackTimestamp = CACurrentMediaTime()
+    setEndTimeViewingImage(time: startFeedbackTimestamp!, isStart: true, type: .feedback)
```
ChaconC, line 264: The calls to `setEndTimeViewingImage` are duplicating the result of the previous line.

Does it make sense to keep the `setEndTimeViewingImage` method if you're setting it directly? It looks like you were trying to use it as a helper method.

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -184,271 +124,257 @@ class GameManager {
...
+
     if gameParameters.showFeedback {
-      delegate?.updateText(text: Constants.timeRespondText, color: .black, font: Constants.smallFont, isStart: false, typeTime: .feedback)
+        delegate?.updateText(text: Constants.timeRespondText, color: .black, font: Constants.smallFont, isStart: false, typeTime: .feedback)
```
ChaconC, line 267: Indentation

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -184,271 +124,257 @@ class GameManager {
...
+        stimulus: text,
+        button_pressed: nil,
+        image_time: endTrialTimestamp! * 1000, // має намалювати
+        correct: false,
+        start_timestamp: 0, // вже намальовано
+        tag: Constants.tag,
+        trial_index: countTest + 1,
+        start_time: startTrialTimestamp * 1000,
+        response_touch_timestamp: 0
+    )
```
ChaconC, line 283: There's a lot of indentation issues in all the changes.

```suggestion
    let model = FlankerModel(
      rt: 0.0,
      stimulus: text,
      button_pressed: nil,
      image_time: endTrialTimestamp! * 1000,  // має намалювати
      correct: false,
      start_timestamp: 0,  // вже намальовано
      tag: Constants.tag,
      trial_index: countTest + 1,
      start_time: startTrialTimestamp * 1000,
      response_touch_timestamp: 0
    )
```

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -184,271 +124,257 @@ class GameManager {
...
+  func handleEndOfGame() {
+    guard let gameParameters = gameParameters else { return }
 
-private extension GameManager {
-  func randomString(length: Int = 5) -> String {
-    let letters = "<>"
-    let lettersArray = Array(letters)
-    let symbolArray = [">>", "<<", "--"]
+    endFeedbackTimestamp = CACurrentMediaTime()
+    setEndTimeViewingImage(time: endFeedbackTimestamp!, isStart: false, type: .feedback)
```
ChaconC, line 300: What's the purpose of updating the timestamp at this stage since we're no longer going to use it?

ChaconC (commented):
Tested and input detection is running great now, I have a few questions about the changes you introduced.

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -184,271 +124,257 @@ class GameManager {
...
-      let respondTouchButton = respondTouchButton
-    else { return }
-    let resultTime = (respondTouchButton - startTrialTimestamp) * 1000
-    arrayTimes.append(resultTime.convertToInt())
+    guard let startTrialTimestamp = startTrialTimestamp else { return }
+    var resultTime = (respondTouchButton! - startTrialTimestamp) * 1000
+
+    if resultTime < 0 {
+        resultTime = 0
+    }
```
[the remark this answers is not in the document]
felipeMetaLab: sorry I didnt understand your question, are you asking about the condition i"f resultTime < 0" ?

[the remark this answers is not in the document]
felipeMetaLab: I can remove it, but it was initially created to avoid an error when the result Time was coming in negative

ios/FlankerNativeComponents/GameManager.swift
felipeMetaLab: I removed startTrialTimestamp in the setEndTimeViewingImage method for case .trial to prevent overwriting the startTrialTimestamp that is set elsewhere in your code. Previously, setting startTrialTimestamp here was causing issues with calculating the average response time because it was unintentionally overwriting the timestamp.

By setting startTrialTimestamp directly where it's needed (in the setText() method), we ensure that the correct timestamp is used for response time calculations. Removing it from setEndTimeViewingImage avoids conflicts and maintains accurate timing.

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -110,70 +108,12 @@ class GameManager {
...
-                                   start_timestamp: 0,
-                                   tag: "feedback",
-                                   trial_index: countTest,
-                                   start_time: startFeedbackTimestamp * 1000,
-                                   response_touch_timestamp: 0)
-          delegate?.resultTest(avrgTime: nil, procentCorrect: nil, data: model, dataArray: nil, isShowResults: gameParameters.showResults, minAccuracy: gameParameters.minimumAccuracy)
-          resultManager.addStepData(data: model)
-        }
       case .trial:
-        startTrialTimestamp = time
```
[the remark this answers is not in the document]
felipeMetaLab: I removed startTrialTimestamp in the setEndTimeViewingImage method for case .trial to prevent overwriting the startTrialTimestamp that is set elsewhere in your code. Previously, setting startTrialTimestamp here was causing issues with calculating the average response time because it was unintentionally overwriting the timestamp.

By setting startTrialTimestamp directly where it's needed (in the setText() method), we ensure that the correct timestamp is used for response time calculations. Removing it from setEndTimeViewingImage avoids conflicts and maintains accurate timing.

ios/FlankerNativeComponents/GameManager.swift
felipeMetaLab: You're correct. The call to setEndTimeViewingImage with type: .response is unnecessary because, in setEndTimeViewingImage. 
I will update it.

ios/FlankerNativeComponents/GameManager.swift
felipeMetaLab: You're correct. Since we modified setEndTimeViewingImage to do nothing when type is .trial and isStart is true, this call is unnecessary.
I will remove it.

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -184,271 +124,257 @@ class GameManager {
...
+          delegate?.updateFixations(image: image, isStart: true, typeTime: .trial)
+      } else {
+          delegate?.updateText(text: text, color: .black, font: Constants.bigFont, isStart: true, typeTime: .trial)
+      }
+
+      DispatchQueue.main.asyncAfter(deadline: .now() + 0.05) {
+          self.delegate?.setEnableButton(isEnable: true)
+          self.timeResponse = Timer(timeInterval: gameParameters.trialDuration / 1000, target: self, selector: #selector(self.timeResponseFailed), userInfo: nil, repeats: false)
+          RunLoop.main.add(self.timeResponse!, forMode: .common)
+      }
```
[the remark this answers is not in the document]
felipeMetaLab: The 0.05 second delay is intended to ensure that the UI has fully updated and rendered the stimulus before enabling user interaction and starting the response timer (there is no fixed parameter for it, I just  tested with several times and found this one the best) but I can remove the delay

ios/FlankerNativeComponents/GameManager.swift
felipeMetaLab: right, Since we're setting the timestamp variables directly before calling setEndTimeViewingImage, and the method also set the same variables, this results is redundant code.

ios/FlankerNativeComponents/GameManager.swift
felipeMetaLab: Initially it was a screen for I check the end of the cycle, but I Kept the screen, I can remove  it.

ios/FlankerNativeComponents/GameManager.swift
felipeMetaLab: I removed.

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -184,271 +124,257 @@ class GameManager {
...
+        tag: Constants.tag,
+        trial_index: countTest + 1,
+        start_time: startTrialTimestamp * 1000,
+        response_touch_timestamp: 0
+    )
 
     resultManager.addStepData(data: model)
-    delegate?.resultTest(avrgTime: nil, procentCorrect: nil, data: model, dataArray: nil, isShowResults: gameParameters.showResults, minAccuracy: gameParameters.minimumAccuracy)
-    responseText = Constants.timeRespondText
+    delegate?.resultTest(avrgTime: nil, procentCorrect: nil, data: model, dataArray: nil, isShowResults: false, minAccuracy: gameParameters.minimumAccuracy)
```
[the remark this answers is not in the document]
felipeMetaLab: yes.

ChaconC (approved):
Thanks @felipeMetaLab 
Tested multiple runs and all taps appear registered correct now.

PR #880, opened 2024-10-28
fix: retaining local data on log out (M2-8113)

qiushihe:
### 📝 Description

🔗 [Jira Ticket M2-8113](https://mindlogger.atlassian.net/browse/M2-8113)

This PR removes the `await cleanupData()` from the "log out" hook.

The call was previously erroneously added by me. I thought it was an oversight that entity record/progression data were not cleared upon log out. I didn't realize retaining those data was an intended behaviour. So this PR removes the call to restore the previously intentional behaviour of not clearing entity record/progression data on log out.

### 📸 Screenshots

N/A

### 🪤 Peer Testing

1. Log into the mobile app
2. Start an activity but do not complete it
3. Exit the activity
4. Log out
5. Re-log into the mobile app
6. Resume the activity

... at this point, the activity should resume from where it was left off before existing and logging out.

### ✏️ Notes

N/A


farmerpaul (approved):
LGTM

PR #884, opened 2024-10-30
fix: add applets refresh after answers submission (M2-8144)

qiushihe:
### 📝 Description

🔗 [Jira Ticket M2-8144](https://mindlogger.atlassian.net/browse/M2-8144)

This PR adds an extra step to refresh applets data after answers are submitted.

This addition is meant to address the issue in [M2-8144](https://mindlogger.atlassian.net/browse/M2-8144), where when an daily activity is completed via notification, the activity does not disappear from the list of activities. In that case, the activity would disappear from the list if the user goes to refresh the applet. So this PR inserts the same refresh call to the end of the answers submission queue processor. So now, regardless if a activity is completed from the activities listing screen, or from notification, the applets will be refreshed after answers submission.

### 📸 Screenshots

N/A

### 🪤 Peer Testing

1. Create an applet
2. Create an activity
3. Set the activity as "daily", and attach a notification
4. Log into the mobile app
5. Navigate to the activities listing screen
6. Exit the mobile app ... and wait for notification to show up
7. Tap no notification
8. Complete the activity

At this point, the just completed activity should no longer appear in the activities listing screen.

### ✏️ Notes

N/A

[M2-8144]: https://mindlogger.atlassian.net/browse/M2-8144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/entities/activity/lib/hooks/useQueueProcessing.ts
+++ b/src/entities/activity/lib/hooks/useQueueProcessing.ts
@@ -4,6 +4,7 @@ import { useForceUpdate } from '@app/shared/lib/hooks/useForceUpdate';
 import { getDefaultChangeQueueObservable } from '@app/shared/lib/observables/changeQueueObservableInstance';
 import { getDefaultUploadObservable } from '@app/shared/lib/observables/uploadObservableInstance';
 import { getDefaultLogger } from '@app/shared/lib/services/loggerInstance';
+import { useRefreshMutation } from '@entities/applet/model/hooks/useRefreshMutation.ts';
 
 import { getDefaultAnswersQueueService } from '../services/answersQueueServiceInstance';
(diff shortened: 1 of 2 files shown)


mbanting (approved):
Verified this fixes the issue. Out of curiosity, I also tested this with an older version, [2.3.1](https://github.com/ChildMindInstitute/mindlogger-app-refactor/releases/tag/v2.3.1) and the issue wasn't there. So this was introduced in [2.4.0](https://github.com/ChildMindInstitute/mindlogger-app-refactor/releases/tag/v2.4.0). Ideally we revisit this and determine what the root cause is (I'll add a task to the backlog), but refreshing can work in the interim. Code LGTM. Thanks @qiushihe

PR #885, opened 2024-10-30
fix: issue with invalidation of applet base info query data (M2-8151)

qiushihe:
### 📝 Description

🔗 [Jira Ticket M2-8151](https://mindlogger.atlassian.net/browse/M2-8151)

This PR fixes an issue introduced in https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/873 where the "base info" query data cache is not invalidated, leading to the app crashing after an activity/flow is added to an existing applet.

### 📸 Screenshots

N/A

### 🪤 Peer Testing

1. Create an applet
2. Create an activity
3. Log into the mobile app
4. Go into the activity listing screen
5. Create another activity
6. In the mobile app, go back to the applet listing screen and refresh
7. Go into the activity listing screen

At this point, the app should not crash.

### ✏️ Notes

N/A

### DIFF
--- a/src/shared/api/services/appletsService.ts
+++ b/src/shared/api/services/appletsService.ts
@@ -10,6 +10,8 @@ import {
   AppletAndActivitiesDetailsResponse,
   AppletAssignmentsRequest,
   AppletAssignmentsResponse,
+  AppletBaseInfoRequest,
+  AppletBaseInfoResponse,
   AppletDetailsRequest,
   AppletDetailsResponse,
   AppletsResponse,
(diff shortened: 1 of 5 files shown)


src/shared/api/services/appletsService.ts
```diff
@@ -10,6 +10,8 @@ import {
   AppletAndActivitiesDetailsResponse,
   AppletAssignmentsRequest,
   AppletAssignmentsResponse,
+  AppletBaseInfoRequest,
+  AppletBaseInfoResponse,
```
qiushihe, line 14: I don't really know how these missing imports passed the original PR. Maybe there is something wrong with the CI running type-checks? 🤔

farmerpaul: You're right! `tsc` is not being run on the mobile app because there's no deployment script (on the web apps, it's being run as a by-product of setting up the Amplify preview environment). Could you fix this in this PR by cherry-picking [this commit](https://github.com/ChildMindInstitute/mindlogger-app-refactor/pull/886/commits/d55ddc76c06204568736e45c3b3891cea2f864d0) into your PR?

mbanting (approved):
Looks good @qiushihe and confirmed crash is no longer happening. Thank you

PR #895, opened 2024-11-13
fix: Fix greek login screen footer elements wrap (M2-8099)

andrevitalb:
### 📝 Description

🔗 [Jira Ticket M2-8099](https://mindlogger.atlassian.net/browse/M2-8099)

A change to the footer's styling was added, so that in case a longer text or more items are added to the footer, the overflow/hiding problem won't be present.

This change won't be noticeable from previous instances for now, since the elements as they are right now don't present that issue:

image

Changes include:

- Updated styling for footer container (inside `LoginScreen.tsx`)

### 📸 Screenshots

| Before (Optional)                      | After                                 |
| -------------------------------------- | ------------------------------------- |
| image | image |

**NOTE: The change won't be noticeable due to the current text for the three footer elements being different.**

### 🪤 Peer Testing

In project's current state, error can't be reproduced, so there's no validation required.

### DIFF
--- a/src/screens/ui/LoginScreen.tsx
+++ b/src/screens/ui/LoginScreen.tsx
@@ -68,6 +68,9 @@ export const LoginScreen: FC = () => {
             jc={isTablet() ? 'space-around' : 'center'}
             mb={isTablet() ? 50 : 40}
             gap={isTablet() ? 0 : 20}
+            width="100%"
+            flexWrap="wrap"
+            ai="center"
           >
             <Link
               textDecorationLine="underline"


felipeMetaLab:
Looks good, but I am wondering, this change has to be applied for all languages or only greek ?

felipeMetaLab (approved):

andrevitalb:
@felipeMetaLab Ideally, all languages would be good with this change. In the event of, let's say a fourth item is added to the footer, or the current texts are updated to something a bit longer, this change would be needed as well.

PR #900, opened 2024-11-20
M2-8255: fix timing mismatch

felipeMetaLab:
### 📝 Description
🔗 [Jira Ticket M2-8255](https://mindlogger.atlassian.net/browse/M2-8255)
The issue was primarily due to the timing mismatch between when the user could interact with the UI and when the app was ready to process input.
reordering the sequence of operations, we align the app readiness with the user ability to interact.

### 📸 Screenshots
ISSUE VIDEO:

https://github.com/user-attachments/assets/c67784e2-987f-4583-ad13-01ea11ecf810

### 🪤 Peer Testing
* lauch the app
* Run task on an iOS device

There were 2 ways to reproduce the issue easily on my side:
* tap the button as many times as possible 
* tap the button as accurate as possible (stimulus is shown for 500 ms)

### ✏️ Notes

Check more information on the comments of this ticket:
https://mindlogger.atlassian.net/browse/M2-7967


ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -239,13 +241,17 @@ class GameManager {
...
+
+    timeResponse = Timer(timeInterval: gameParameters.trialDuration / 1000,
+    target: self,
+    selector: #selector(timeResponseFailed),
+    userInfo: nil,
+    repeats: false)
+    RunLoop.main.add(timeResponse!, forMode: .common)
   }
 
+
```
ChaconC, line 254: Can you please fix formatting in these lines

ios/FlankerNativeComponents/GameManager.swift
```diff
@@ -137,14 +137,14 @@ class GameManager {
 
   func checkedAnswer(button: SelectedButton) {
     guard !hasRespondedInCurrentTrial else { return }
+    guard let startTrialTimestamp = startTrialTimestamp else { return }
```
ChaconC, line 140: With this change I'm encountering a `index out of range` error in `checkAnswer` at line 157
`let correctChoice = gameParameters.trials[countTest].correctChoice`

CleanShot 2024-11-22 at 09 15 56

felipeMetaLab: Could you test again ? I built several times and I didnt get this error but even so I added a line to prevent this error. 
let me know if it is working.

ChaconC: I was unable to reproduce, but I see the fix you added and makes sense for the issue I encountered

ChaconC (approved):

PR #905, opened 2024-11-28
feat: Add mobile app activity group empty states (M2-8098)

farmerpaul:
### 📝 Description

🔗 [Jira Ticket M2-8098](https://mindlogger.atlassian.net/browse/M2-8098)

Add empty states to the activities list in the Mobile App.

**No activities available:**

**Only scheduled activities available:**

### 🪤 Peer Testing

1. Create an applet with only 1 activity whose auto-assign setting is **turned off** (so no assignments present).
2. Refresh the Mobile App and open that applet.
    **Expected outcome:** The full-screen empty state should be shown
3. Create a new activity in that applet with auto-assign **turned on**, and schedule that activity to be available today sometime after the current time.
4. Refresh the Mobile App and open that applet.
    **Expected outcome:** The "only scheduled activities available" empty state should be shown


ChaconC (approved):

mbanting (approved):
nice work!

ios/MindloggerMobile.xcodeproj/project.pbxproj
mbanting, line 1: Weirdly enough, on my laptop, it changes it back to 
`OTHER_LDFLAGS = "$(inherited)";`
 but ok with making this change especially if this is the case for most on the team. 

Just in case, here are the details of my local RN environment
MacOS: 15.0.1
XCode: 16
CocoaPods: 1.12.1

PR #912, opened 2024-12-03
Fix: M2-8294 Single selection per row: The user remains on the input screen after selecting the response and tapping 'Next'

felipeMetaLab:
### 📝 Description
🔗 [Jira Ticket M2-8294](https://mindlogger.atlassian.net/browse/M2-8294)
Fixing how singleSelectionPerRow get value

### 📸 Screenshots
Screenshot 2024-12-02 at 11 48 13 PM
Screenshot 2024-12-02 at 11 46 32 PM
Screenshot 2024-12-02 at 11 48 19 PM

https://github.com/user-attachments/assets/1e261093-663f-4a8c-8315-dce9bd235681

### 🪤 Peer Testing

1 - create an activity with  single selection per row Item
2 - create a conditional logic (could be any conditional logic as long it is is equal or it is not equal) 
3 - launch the app 
4 - select the applet with the activity you created and check if the logic is responding right.

### DIFF
--- a/src/features/pass-survey/model/AnswerValidator.ts
+++ b/src/features/pass-survey/model/AnswerValidator.ts
@@ -112,6 +112,19 @@ export function AnswerValidator(
     return null;
   };
 
+  const getSelectionRowOptionValue = (rowIndex: number): string | null => {
+    const answer = currentAnswer?.answer as Maybe<
+      { id: string; text: string }[]
+    >;
+
+    if (answer && answer[rowIndex]) {
+      const optionIds = answer[rowIndex].id;
+
+      return optionIds;
+    }
+    return null;
+  };
+
   return {
     isCorrect() {
       if (!currentPipelineItem?.validationOptions) {
@@ -164,12 +177,12 @@ export function AnswerValidator(
     },
 
     isEqualToRowOption(rowIndex: number, optionValue: string) {
-      const selectedOption = getRowOptionValues(rowIndex);
+      const selectedOption = getSelectionRowOptionValue(rowIndex);
       return selectedOption !== null && selectedOption.includes(optionValue);
     },
 
     isNotEqualToRowOption(rowIndex: number, optionValue: string) {
-      const selectedOption = getRowOptionValues(rowIndex);
+      const selectedOption = getSelectionRowOptionValue(rowIndex);
       return selectedOption !== null && !selectedOption.includes(optionValue);
     },
 


src/features/pass-survey/model/AnswerValidator.ts
```diff
@@ -164,12 +177,12 @@ export function AnswerValidator(
...
 
     isEqualToRowOption(rowIndex: number, optionValue: string) {
-      const selectedOption = getRowOptionValues(rowIndex);
+      const selectedOption = getSelectionRowOptionValue(rowIndex);
       return selectedOption !== null && selectedOption.includes(optionValue);
     },
 
     isNotEqualToRowOption(rowIndex: number, optionValue: string) {
-      const selectedOption = getRowOptionValues(rowIndex);
+      const selectedOption = getSelectionRowOptionValue(rowIndex);
```
ChaconC, line 185: Ran into an issue here, if the value is null it doesn't pass this validation (`value != option`, which `null !== option`)

* Single Select Item configured as a 2-item matrix

CleanShot 2024-12-03 at 21 11 11

* Condition says:
  * if first row = o1, show message1
  * if second row != o1, show message2

CleanShot 2024-12-03 at 21 09 58

Video:
Second row is null and it fails validation, even though its allowed by the item type

https://github.com/user-attachments/assets/b0f047b5-6dc6-4c59-bb1e-accb8718a180

ChaconC (changes requested):

src/features/pass-survey/model/AnswerValidator.ts
felipeMetaLab: It  looks right for me, for example, if we enable the null as a right answer, it means every condition will pass as long you choose only one option.

Should I considerer null as valid answer ?

felipeMetaLab: in my understanding the logic is like: if the row 2 of item "single is selected turn it visible, otherwise keep it invisible, turn null into right answer, we are saying: "ok if dont select we are considerer you want to see it"

felipeMetaLab: its looks right for me, because we are not accepting null as a right answer.

Example, I understand the logic is like: ok if you select something different of row2 - o1 it will appears, but if we consider empty or null as a right value, we are saying: "ok, even if you dont select nothing, we are considering your answer as different of conditional option"

felipeMetaLab: let me know your thoughts, should we enable null as different in terms of conditional logic comparison  ?

felipeMetaLab: if yes, it is fast to solve.

ChaconC (approved):

PR #927, opened 2025-01-14
Fix:M2-8467 resolving disparity between release 2.7 and dev

felipeMetaLab:
### 📝 Description

🔗 [Jira Ticket M2-8467](https://mindlogger.atlassian.net/browse/M2-8467)

Resolving disparity between dev and release 2.7

### 🪤 Peer Testing

1 - launch the app
2 - check if everything is going right
3 - at this point the main change was the dependencies, so the most important is checking the requests and all related functionality.


felipeMetaLab:
@ChaconC  the dev contains some things that the release does not contain, in order to resolve the disparity, I recommend to merge, otherwise, we will have many things to equalize or resolving.
but its possible to mitigate and resolve after.

ChaconC:
@felipeMetaLab `release/2.7.0` was released to prod already, this PR is not needed.

felipeMetaLab:
ok, so. this PR doesnt make sense.

PR #932, opened 2025-02-07
feat: Update mobile app to send device properties on each request (M2-8664)

sultanofcardio:
### 📝 Description

🔗 [Jira Ticket M2-8664](https://mindlogger.atlassian.net/browse/M2-8664)

This PR updates the Axios interceptor to add the following HTTP headers to each request:
- **Device-Id**: The device ID currently sent on sign up or login
- **OS-Name**: `android` or `ios`
- **OS-Version**: The version of the OS (e.g. 13 for Android 13)
- **App-Version**: The version of the MindLogger app currently installed.

### 📸 Screenshots

When paired with https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1727, the `user_devices` table will now populate these extra properties instead of leaving them blank. This happens on login, so you can use that behaviour to confirm it works

image

### 🪤 Peer Testing

Using https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1727, run the API and log into the MindLogger app. Then inspect the `user_devices` table in the database.

Alternatively you can print the headers on any random request in the API code.

### ✏️ Notes

Despite references to https://github.com/ChildMindInstitute/mindlogger-backend-refactor/pull/1727, the two PRs are not dependent on each other and can be deployed independently

### DIFF
--- a/src/jobs/request-interception.ts
+++ b/src/jobs/request-interception.ts
@@ -1,9 +1,15 @@
+import { Platform, PlatformIOSStatic } from 'react-native';
+
 import i18n from 'i18next';
+import { PlatformAndroidStatic } from 'react-native/Libraries/Utilities/Platform';
 
 import { getDefaultSessionService } from '@app/entities/session/lib/sessionServiceInstance';
 import { httpService } from '@app/shared/api/services/httpService';
 import { TIMEZONE } from '@app/shared/lib/constants/dateTime';
 import { createJob } from '@app/shared/lib/services/jobManagement';
+import { APP_VERSION } from '@shared/lib/constants';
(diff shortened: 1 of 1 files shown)


ChaconC (approved):
Works as expected and I can see the headers in the request.
This is more of a BE issue but a quick note that `UserDeviceCreate` is only called on `/login`, meaning if the user is already logged in the data never gets stored.

sultanofcardio:
Thanks for the tip @ChaconC. You're right, this ticket is really just for sending the properties. The BE only needs to store them on sign up and login for now, but may optionally choose to use them while processing a request in the future 👌

PR #951, opened 2025-03-04
fix: correct Lato assets, default `lineHeight` (M2-8638)

farmerpaul:
### 📝 Description

🔗 [Jira Ticket M2-8638](https://mindlogger.atlassian.net/browse/M2-8638)

The wrong Lato font assets were being used in the previous implementation; this change updates them.
Also, a default `lineHeight` needs to be explicit throughout the app as Android and iOS make different assumptions for a reasonable default value. Chose 125% as the default, which is overridable if provided.

### 📸 Screenshots

The three columns are:
1. Android – with these changes
2. iOS – previously
3. iOS – with these changes

CleanShot 2025-03-04 at 14 19 50@2x

### 🪤 Peer Testing

To test, you'll need to **rebuild the app**. For iOS, I recommend running `yarn clean-ios` then `yarn ios`.

Test on both iOS and Android.

To confirm it's working:
- Check that the correct font and line height is being used for UI when the Greek language is selected.
- Confirm that markdown areas (like the About Mindlogger screen and markdown contained in assessments) are also styled correctly.


andrevitalb (approved):

mbanting (approved):
Thanks for sorting out the visual differences and behaviour between platforms

PR #952, opened 2025-03-05
fix: Mobile app fixes (M2-8800, M2-8801)

farmerpaul:
### 📝 Description

🔗 [Jira Ticket M2-8800](https://mindlogger.atlassian.net/browse/M2-8800)
🔗 [Jira Ticket M2-8801](https://mindlogger.atlassian.net/browse/M2-8801)

The "Processing Answers" text was being cut off after the recent line height change; this was related to shorthand prop names complicating the default line height calculation.

Also, using the `headerTitle` component affected the default value for `headerBackVisible` on Android only. Fixed by setting it explicitly to `false`.

### 📸 Screenshots

| Before                      | After                                 | |
| -------------------------------------- | ------------------------------------- | - |
| processing answers trim | Screenshot_1741182034 |  |
| extra back button on android |  Screenshot_1741181840 | Screenshot_1741181884 |

### 🪤 Peer Testing

1. Submit an assessment and observe the Processing Answers screen
    **Expected outcome:** "Processing Answers" should not be cut off.
2. In Android, navigate to various subscreens, including "About Mindlogger", "Create Account", "User Settings", etc.
    **Expected outcome:** There should only be 1 back button.

### DIFF
--- a/src/shared/ui/Text.tsx
+++ b/src/shared/ui/Text.tsx
@@ -1,6 +1,6 @@
 import { Text as TGUIText, TextProps } from '@tamagui/core';
 
-type Props = TextProps;
+type Props = Omit<TextProps, 'fos' | 'fow' | 'lh'>;
 
 export const Text = ({
   fontSize = 16,
--- a/src/screens/config/theme.tsx
+++ b/src/screens/config/theme.tsx
@@ -34,6 +34,7 @@ export const getScreenOptions = ({
     },
     headerShadowVisible: false,
     headerTitle: HeaderTitle,
+    headerBackVisible: false,
     headerLeft: () => (
       <Text
         accessibilityLabel="close-button"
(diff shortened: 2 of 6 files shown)


src/shared/ui/Text.tsx
```diff
@@ -1,6 +1,6 @@
 import { Text as TGUIText, TextProps } from '@tamagui/core';
 
-type Props = TextProps;
+type Props = Omit<TextProps, 'fos' | 'fow' | 'lh'>;
```
andrevitalb, line 3: Nice safeguard 🤓

andrevitalb (approved):
Can confirm issues are gone with these changes. Hadn't seen the second way of applying font-related styling. Kind of doesn't make much sense to have two sepparate ways of changing the same thing, but, oh well 🤷🏻‍♂️

PR #953, opened 2025-03-05
fix: Add Lato italic font (M2-8638)

farmerpaul:
### 📝 Description

🔗 [Jira Ticket M2-8638](https://mindlogger.atlassian.net/browse/M2-8638)

No italicized version of Lato was included when it was added to support Greek font rendering. This PR adds support for it.

The only place that italicized text an appear is in markdown content contained in activity descriptions.

> [!NOTE]
> For the font to be properly applied, it needs to be specified in custom styles in the `MarkdownView` component. However, the markdown library we're using has a primitive styling system with no way to specify the font for text that is **_both bold and italics_**. As it turns out, Android chooses to display such text as **bold**, while iOS choose to display such text as _italics_. This is strictly an issue with the Lato font, when Greek language is selected, and not with the default font.
>
> Confirmed with @jodybrookover that it is not necessary to properly style bold + italics text for CAMHI.

### 📸 Screenshots

**Default font (when non-Greek language is selected):**

CleanShot 2025-03-05 at 14 34 34@2x

**Lato font (when Greek language is selected):**

CleanShot 2025-03-05 at 14 30 13@2x

### 🪤 Peer Testing

Test in both iOS and Android:

- Create an applet with an activity containing an item whose description contains italicized text.
- In the mobile app, set the language to Greek, and perform an assessment of the activity in the mobile app.
    **Expected outcome:** Italicized text is rendered using the italicized Lato font


ChaconC (approved):

mbanting (approved):
Code looks good
Peer testing on iOS successfully
Unfortunately running to local build problems for Android (nothing to do with this PR). Don't want to hold it up any longer, the risk is low and QA will test anyways. Approving now. Thanks @farmerpaul!

PR #1018, opened 2025-08-13
fix: prevent drawing grid cutoff when 'Evenly Distribute' is disabled

maxto024:

  ### 📝 Description

  
  

  🔗 [Jira Ticket M2-9627](https://mindlogger.atlassian.net/browse/M2-9627)

  This PR fixes a critical UI issue where the drawing grid's last row was being cut off when the "Evenly Distribute the Drawing Example and Drawing Canvas Area" option was disabled. This forced users to scroll during timed assessments,
  preventing them from seeing the complete grid and key/legend simultaneously.

  Changes include:

  - Dynamic image sizing based on actual aspect ratio instead of hardcoded 300×300 dimensions
  - Device-specific safety margins (94px for tablets, 74px for phones) to account for UI elements
  - Smart space distribution algorithm that balances example image and canvas sizes
  - Minimum canvas visibility guarantee (60% of width or 300px, whichever is smaller)
  - Removed unnecessary `async` keyword from `onResult` function to resolve linting warning

  ### 📸 Screenshots

  | Before                                     | After                                    |
  | ------------------------------------------ | ---------------------------------------- |
  | Grid cut off - requires scrolling | Full grid visible - no scrolling |
  | Last row of drawing grid hidden below fold | Complete grid visible with key/legend    |

  ### 🪤 Peer Testing

  **Requires `yarn install`**
  **Requires `yarn pods`**

  1. Navigate to a drawing assessment item with the "Evenly Distribute" option **disabled**

     **Expected outcome:** Drawing item should load with both example image and canvas visible

  2. Verify the drawing grid is fully visible without scrolling

     **Expected outcome:** All rows of the drawing grid should be visible, including the last row with the key/legend

  3. Test on various devices (phones and tablets)

     **Expected outcome:** Layout should adapt appropriately with no cut-off content on any device size

  4. Test with different image aspect ratios (portrait, landscape, square)

     **Expected outcome:** Both the example image and canvas should resize dynamically while maintaining visibility

  5. Verify drawing functionality still works correctly

     **Expected outcome:** Users can draw on the canvas as before, with all drawing tools functional

  ### ✏️ Notes

  - This fix only affects `DrawingTestLegacyLayout.tsx` which is used when `proportionEnabled` is false
  - The solution uses the `useImageDimensions` hook to get actual image dimensions for proper scaling
  - Safety margins were tested on various iOS devices including iPads and iPhones
  - The smart space distribution prioritizes canvas visibility while maximizing example image size

### DIFF
--- a/src/entities/drawer/ui/DrawingTest/DrawingTestLegacyLayout.tsx
+++ b/src/entities/drawer/ui/DrawingTest/DrawingTestLegacyLayout.tsx
@@ -3,6 +3,7 @@ import { FC } from 'react';
 
 import { CachedImage } from '@georstat/react-native-image-cache';
 
+import { useImageDimensions } from '@app/shared/lib/hooks/useImageDimensions';
 import { Box, XStack } from '@app/shared/ui/base';
 
 import { DrawingTestProps } from './DrawingTest';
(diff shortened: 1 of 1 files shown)


src/shared/lib/featureFlags/FeatureFlagsService.ts
```diff
@@ -61,6 +61,18 @@ export class FeatureFlagsService implements IFeatureFlagsService {
...
+    // Dev environment overrides
+    if (__DEV__) {
+      const devOverrides: Record<string, boolean> = {
+        'enable-better-drawing-image-sizing': true,
+        // Add other feature flags here as needed for dev environment
+      };
+
+      if (flag in devOverrides) {
+        return devOverrides[flag];
+      }
```
andrevitalb, line 73: Instead of doing this approach, can we do something similar to the other repos in which we have default values for each of the flags? This way we don't rely on manual overrides for each env, since this would cause inconsistencies in envs such as UAT, for example.

Take [admin repo's implementation](https://github.com/ChildMindInstitute/mindlogger-admin/blob/develop/src/shared/hooks/useFeatureFlags.const.ts) as example.

src/entities/drawer/ui/DrawingTest/DrawingTestLegacyLayout.tsx
```diff
@@ -29,7 +43,86 @@ export const DrawingTestLegacyLayout: FC<DrawingTestProps> = props => {
     props.onResult(result);
   };
 
-  return (
+  // Calculate dimensions based on feature flag
+  const dimensions = useMemo<LayoutDimensions>(() => {
+    if (!enableBetterDrawingImageSizing) {
+      // OLD IMPLEMENTATION: Fixed 300x300 sizing
+      return {
+        exampleImageWidth: 300,
```
andrevitalb, line 51: For all these numbers that act as constants, can we define them as such at the top of this file?

```ts
const RECT_PADDING = 15;
const DEFAULT_IMAGE_SIZE = 300;
const GAP_SIZE = 20;
const MIN_EXAMPLE_HEIGHT = 50;
const FALLBACK_EXAMPLE_HEIGHT_FACTOR = 0.25;
const MAX_FALLBACK_EXAMPLE_HEIGHT = 150;
const MIN_CANVAS_SIZE_FACTOR = 0.6;
const MIN_CANVAS_SIZE = 300;
```

src/entities/drawer/ui/DrawingTest/DrawingTestLegacyLayout.tsx
```diff
@@ -29,7 +43,86 @@ export const DrawingTestLegacyLayout: FC<DrawingTestProps> = props => {
...
+    }
+
+    const remainingHeight = availableHeight - exampleImageHeight - gap;
+    const canvasSize = Math.min(availableWidth, remainingHeight);
+
+    return {
+      exampleImageWidth,
+      exampleImageHeight,
+      canvasSize,
+    };
```
andrevitalb, line 115: I believe this can be broken down into a couple helper functions that enforce separation of concerns and aid in readability and maintainability in the longer run.

For example:

```ts
interface SafetyMargins {
  bottomMargin: number;
  additionalMargin: number;
}

const calculateSafetyMargins = (isTablet: boolean): SafetyMargins => ({
  bottomMargin: 24, // mb="$6"
  additionalMargin: isTablet ? 70 : 50,
});
```

```ts
const calculateAvailableSpace = (
  containerWidth: number,
  containerHeight: number,
  isTablet: boolean,
): { availableWidth: number; availableHeight: number } => {
  const { bottomMargin, additionalMargin } = calculateSafetyMargins(isTablet);
  const safetyMargin = bottomMargin + additionalMargin;
  
  return {
    availableWidth: containerWidth,
    availableHeight: containerHeight - safetyMargin,
  };
};
```

src/entities/drawer/ui/DrawingTest/DrawingTestLegacyLayout.tsx
```diff
@@ -29,7 +43,86 @@ export const DrawingTestLegacyLayout: FC<DrawingTestProps> = props => {
...
+            availableWidth / exampleImageDimensions.aspectRatio,
+          );
+        }
+      } else {
+        exampleImageHeight = Math.min(availableHeight * 0.25, 150);
+        exampleImageWidth = Math.floor(
+          exampleImageHeight * exampleImageDimensions.aspectRatio,
+        );
+      }
+    }
```
andrevitalb, line 106: Adding a couple comments on the specific sizing cases for better understanding can be helpful for anyone who reads this.

andrevitalb (changes requested):
Updates to `DrawingTestNewLayout.tsx` would be missing as well

src/entities/drawer/ui/DrawingTest/DrawingTestLegacyLayout.tsx
maxto024: Done

src/entities/drawer/ui/DrawingTest/DrawingTestLegacyLayout.tsx
maxto024: done added

maxto024:
> Does this issue not affect `DrawingTestNewLayout.tsx` as well?

Yes it doesn't, we only use that when we have ['Evenly Distribute' is enabled]

src/entities/drawer/ui/DrawingTest/DrawingTestLegacyLayout.tsx
maxto024: **broken down the complex dimension calculation logic into several focused helper functions:**

-   1. calculateSafetyMargins - Determines the bottom margin and additional margin based on device type
-   2. calculateAvailableSpace - Calculates the available width and height after accounting for safety margins
-   3. calculateInitialExampleImageSize - Calculates the initial example image dimensions based on available width
-   4. calculateMinCanvasSize - Determines the minimum canvas size with appropriate constraints
-   5. shrinkImageToFit - Handles the complex logic of shrinking the image when it doesn't fit vertically
-   6. calculateFinalCanvasSize - Calculates the final canvas size based on remaining spac

PR #1025, opened 2025-08-27
fix:spanish about page - minor heading fix (M2-9696)

jodybrookover:
### 📝 Description

🔗 [Jira Ticket M2-9696](https://mindlogger.atlassian.net/browse/M2-9696)

Had a minor issue with the heading of the spanish page containing the first paragraph as well.  

### 🪤 Peer Testing

Change your language to Spanish in the app.
Navigate to the about page
Compare to english to ensure line breaks are the same or at least very close to reasonable


maxto024 (approved):
looks good

PR #1084, opened 2026-04-02
fix: Update `fmt` to build with Xcode 26.4

divbzero:
### 📝 Description

While implementing M2-10032 I came across an issue with `yarn pods:reinstall` failing with Xcode 26.4. Updating the `fmt` dependency `11.0.2` → `12.1.0` resolves the issue. 

### ✏️ Notes

We are patching `react-native@0.79` because it pins `fmt@11.0.2`. We can remove the patch after we upgrade to `react-native@0.85` which incorporates the fix. See [patches/README.md](https://github.com/ChildMindInstitute/mindlogger-app-refactor/blob/M2-10032-update-fmt/patches/README.md) for details.


adeiji (approved):
Looks good!

PR #1107, opened 2026-04-28
fix: update project for Xcode 26 compatibility

adeiji:
### 📝 Description
Updates the project for Xcode 26 compatibility.

🔗 [Jira Ticket M2-#10686](https://mindlogger.atlassian.net/browse/M2-10686)

Xcode 26 introduces a stricter C++ compiler that fails on `consteval` usage in the fmt library (a transitive CocoaPods dependency). The build fails because fmt's `base.h` sets `FMT_USE_CONSTEVAL 1` when `__cpp_consteval` is defined, but Xcode 26's compiler rejects the resulting consteval expressions.

This PR adds a post-install hook in the Podfile that patches `fmt/include/fmt/base.h` to set `FMT_USE_CONSTEVAL 0`, disabling consteval and falling back to constexpr instead. The patch is conditional — it only applies if the fmt pod exists and hasn't already been patched — so it has no effect on earlier Xcode versions.

Changes include:

- Add post-install patch in Podfile to disable consteval in fmt
- Updated Podfile.lock checksums from pod install

### 🪤 Peer Testing
- [ ] Run `yarn pods`
- [ ] Run `yarn ios`
- [ ] Verify iOS build succeeds on Xcode 26
- [ ] Verify smoke test passes

### DIFF
--- a/ios/Podfile
+++ b/ios/Podfile
@@ -70,6 +70,22 @@ abstract_target 'MindloggerMobileCommonPods' do
       # :ccache_enabled => true
     )
 
+    # Xcode 26 workaround: patch fmt to disable consteval
+    fmt_base = File.join(installer.sandbox.root, 'fmt', 'include', 'fmt', 'base.h')
+    if File.exist?(fmt_base)
+      content = File.read(fmt_base)
+      unless content.include?('Xcode 26 workaround')
+        patched = content.gsub(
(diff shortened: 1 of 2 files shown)


divbzero:
This is an alternative to the fix in #1084. @adeiji confirmed that the fix in #1084 works for him too so we’ll stick with that for now.

PR #1114, opened 2026-04-30
fix: unity view now shows full screen

adeiji:
### Description
The Banner component was rendered at the top of the screen at all times and when rendered it does not allow any UI components to be visible at the top of the screen. Added a bannersHidden flag to the banner Redux slice so screens can hide it. UnityView now dispatches setBannersHidden on mount/unmount to remove the bar and allow full-screen rendering

🔗 [Jira Ticket M2-#10635](https://mindlogger.atlassian.net/browse/M2-10635)

Changes include
- Added `bannersHidden` state to the banner Redux slice so any component can request the banner bar be removed from the view tree
- `UnityView` dispatches `setBannersHidden(true)` on mount and `setBannersHidden(false)` on unmount, allowing it to render edge-to-edge on iOS

## Peer testing
- [ ] Open a Unity activity on a physical phone device
- [ ] Verify the Unity view renders into the notch area with no bar covering it
- [ ] Navigate back from the Unity activity and verify the banner bar returns on other screens

### 📸 Screenshots

Screenshot 2026-04-30 at 1 31 38 PM

### ✏️ Notes
Changes still need to be tested on Android.

### DIFF
--- a/src/entities/banner/model/selectors.ts
+++ b/src/entities/banner/model/selectors.ts
@@ -11,3 +11,8 @@ export const bannersBgSelector = createSelector(
   selectBanners,
   banners => banners.bannersBg,
 );
+
+export const bannersHiddenSelector = createSelector(
+  selectBanners,
+  banners => banners.bannersHidden,
+);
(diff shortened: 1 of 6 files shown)


divbzero (approved):
Looks good!

PR #1117, opened 2026-05-08
fix: Patch `react-native-unity` for Android build

divbzero:
### 📝 Description

1. `yarn install --ignore-scripts` for a clean `@azesmway/react-native-unity` install.
2. Update `UPlayer.java` in `node_modules/@azesmway/react-native-unity` as described by @TwigAy-CMI.
3. Run `yarn patch-package @azesmway/react-native-unity`

### ✏️ Notes

The patch addresses an Android build issue that @TwigAy-CMI investigated.

See related discussion in azesmway/react-native-unity#123.


divbzero:
The patch does not work well. We should stick with setting `BUILD_WITH_UNITY` and relying on [postinstall.sh](https://github.com/ChildMindInstitute/mindlogger-app-refactor/blob/2026.04.3/script/postinstall.sh).

PR #1119, opened 2026-05-19
fix: Unity mass data upload (M2-10719)

divbzero:
### 📝 Description

🔗 [Jira Ticket M2-10719](https://mindlogger.atlassian.net/browse/M2-10719)

Changes include:

- Upload Unity media files to S3 with `react-native-blob-util` instead of `axios`. 
- Increment progress indicator for each media file uploaded.

### ✏️ Notes

This should improve performance but not fully verified yet.


adeiji (approved):
Everything looks good to me.

PR #1121, opened 2026-05-22
fix: Unity file path encoding (M2-10719)

divbzero:
🔗 [Jira Ticket M2-10719](https://mindlogger.atlassian.net/browse/M2-10719)

Pull request #1119 resolved the issue in iOS but not in Android. The logs for Android indicated that the Unity files were saved successfully to the file system but could not be uploaded. Turns out that Android tries to URI decode file URIs (prefixed with `file://` scheme) which breaks if there is a literal `%` character in the file path. Consistently passing in the file path (no `file://` scheme) when interacting with the file system resolves the issue.

Changes include:

- Rename `getFilePath` → `getFileUri`. This helper function was previously misnamed.
- Define new `getFilePath` helper function that strips `file://` scheme from file URI.
- Use file path consistently when interacting with the file system.

### DIFF
--- a/src/entities/activity/lib/services/AnswersUploadService.ts
+++ b/src/entities/activity/lib/services/AnswersUploadService.ts
@@ -22,7 +22,7 @@ import {
   formatToDtoDate,
   formatToDtoTime,
 } from '@app/shared/lib/utils/dateTime';
-import { isLocalFileUrl } from '@app/shared/lib/utils/file';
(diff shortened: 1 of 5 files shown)


adeiji (approved):
The changes look good @divbzero

sricharan-varanasi (approved):
Looks good, Chris , Approved. @divbzero

PR #1127, opened 2026-06-02
fix: update `Podfile.lock` for Firebase 21/iOS SDK 11 compatibility (M2-10783)

divbzero:
### 📝 Description

Due to a package update to Firebase v21 other packages we're tied to incorrect versions in `Podfile.lock`. This PR updates the packages to all work with Firebase v21

🔗 [Jira Ticket M2-10783](https://mindlogger.atlassian.net/browse/M2-10783)

### ✏️ Notes

This merges the fix from #1122 into `dev`.


adeiji (approved):
Looks good!

PR #1128, opened 2026-06-03
fix: add missing Xhosa weekday and fall back to English (M2-10835)

sricharan-varanasi:

### 📝 Description

🔗 [Jira Ticket M2-10835](https://mindlogger.atlassian.net/browse/M2-10835)

App crashed on opening an applet when the language was set to Xhosa (`Cannot read property 'toUpperCase' of undefined` in `HorizontalCalendar`). Root cause was a data issue : the Xhosa `calendar.weekdays` string was missing Thursday (`Lwesine`), so it had 6 days instead of 7 and the calendar hit an undefined entry for Saturday.

Changes include:

- Added the missing Xhosa weekday (`Lwesine`) to `xh.json`.
- Added a per-day fallback to English in `HorizontalCalendar` so a missing/malformed translation renders the English label instead of crashing.

### 🪤 Peer Testing

- Switch language to **Xhosa**, log in, and tap on an applet card.

    **Expected outcome:** The applet screen opens normally (no crash) and the calendar shows all 7 weekdays.

### ✏️ Notes

- Affected Xhosa only - Zulu and Afrikaans already had all 7 weekdays.

### DIFF
--- a/assets/translations/xh.json
+++ b/assets/translations/xh.json
@@ -71,7 +71,7 @@
   },
   "calendar": {
     "months": "NgoJanuwari_Februwari_Matshi_Apreli_Meyi_Juni_Julayi_Agasti_Septemba_Okthobha_Novemba_Disemba",
-    "weekdays": "Langa_Mso_Lwesibini_Lwesithathu_Lwesihlanu_Sat",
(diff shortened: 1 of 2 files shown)


divbzero (approved):
Looks good! The fallback is a good idea too.

PR #1129, opened 2026-06-03
fix: show English as 'English' in Zulu language list (M2-10837)

sricharan-varanasi:

### 📝 Description

🔗 [Jira Ticket M2-10837](https://mindlogger.atlassian.net/browse/M2-10837)

English was shown as "IsiNgisi" in the language list when the app language was set to Zulu. The Zulu translation file had translated the English label instead of keeping its native name. Fixed so English always displays as "English", consistent with the other languages which use their native names.

Changes include:

- Updated the Zulu translation (`zu.json`) so the English language label is "English" instead of "IsiNgisi".

### ✏️ Notes

- Web has the same issue, fixed separately (M2-10838 / equivalent).

### DIFF
--- a/assets/translations/zu.json
+++ b/assets/translations/zu.json
@@ -274,7 +274,7 @@
   "language_screen": {
     "app_language": "Ulimi lwe-application",
     "change_app_language": "Shintsha Ulimi",
-    "en": "IsiNgisi",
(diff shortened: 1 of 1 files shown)


adeiji (approved):
Looks good!

PR #1140, opened 2026-06-05
fix: Lock iOS to portrait after app mounts (M2-10845, M2-10877)

divbzero:
### 📝 Description

🔗 [Jira Ticket M2-10845](https://mindlogger.atlassian.net/browse/M2-10845)
🔗 [Jira Ticket M2-10877](https://mindlogger.atlassian.net/browse/M2-10877)

Previously, I tried to lock to portrait when the module loaded which caused the `lockTo` method to be a no-op. Instead, we need to: (a) wait until the app mounts to lock to portrait and (b) override the view controller to constrain orientations because we set a static orientation mask on the app in #1116 to address M2-10056.

### ✏️ Notes

This issue only appeared in iOS because we lock to portrait with the manifest on Android.

### DIFF
--- a/src/app/index.tsx
+++ b/src/app/index.tsx
@@ -1,3 +1,4 @@
+import { useEffect } from 'react';
 import { LogBox, StyleSheet } from 'react-native';
 
 import {
@@ -7,6 +8,9 @@ import {
   SdkVerbosity,
   UploadFrequency,
 } from '@datadog/mobile-react-native';
+import RNOrientationDirector, {
+  Orientation,
+} from 'react-native-orientation-director';
 import Animated, { LinearTransition } from 'react-native-reanimated';
 
 import { Banners } from '@app/entities/banner/ui/Banners';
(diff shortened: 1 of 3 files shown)


src/app/index.tsx
```diff
@@ -70,6 +74,11 @@ if (__DEV__) {
 }
 
 const App = () => {
+  useEffect(() => {
+    // Lock to portrait after app mounts
+    RNOrientationDirector.lockTo(Orientation.portrait);
```
adeiji, line 79: This looks good. One small thing I’m wondering about: if the user has their phone in landscape before opening the app, could there be a brief flash where the app appears in landscape before it locks to portrait? Not a blocker, but worth being aware of.

adeiji (approved):

src/app/index.tsx
divbzero: @adeiji I think you’re right. This flash could in theory happen though I can’t think of a way around it at the moment.

divbzero: @adeiji When I tested it appears that the app displays a white screen as it first loads so I did not observe a flash.

adeiji (approved):
Yes, this is a good approach to solving this issue. Thank you!

PR #1142, opened 2026-06-09
Fix: abtrails android flicker (M2-10772)

sricharan-varanasi:

### 📝 Description

🔗 [Jira Ticket M2-10772](https://mindlogger.atlassian.net/browse/M2-10772)

A/B Trails lines were blinking on Android (Samsung S23 FE, Pixel 9 Pro XL) after connecting the last point or drawing an incorrect line. 

**Root cause:** The drawing canvas (`SketchCanvas`) was being unmounted every time a line was committed or rejected via `hideSketchCanvas()`. On `@shopify/react-native-skia` 2.4.21 (bumped in the 16KB page-size compliance update), Android forces a full repaint on any neighbouring native Skia surface when one is destroyed. That repaint drops one frame : the blink.

**Fix:** Keep `SketchCanvas` always mounted. Instead of unmounting it, disable the drawing gesture via a new `enabled` prop. The canvas stays alive on screen; only touch input is paused. Android sees no surface change, so no forced repaint, no blink.

Changes include:

- `useDrawingGesture.ts` - added `enabled` prop; wires `.enabled(false/true)` to all gesture recognisers (iOS manual + Android pan/tap/longPress)
- `SketchCanvas.tsx` - added `enabled` prop; passes it through to `useDrawingGesture`
- `AbCanvas.tsx` - replaced the `isSketchCanvasShown && !readonly` mount gate with `enabled={isSketchCanvasShown && !readonly}` on an always-mounted `SketchCanvas`; null `currentPathRef` instead of calling `resetCurrentPath()` in the completion and wrong-line branches to avoid clearing the live stroke at the same instant React is painting a new committed/error line; clear the live stroke in the error timeout instead, where no concurrent paint is happening

### 🪤 Peer Testing

- Open an A/B Trails activity on an Android device.

- Connect points in the correct order up to the last connection. Complete the trail by connecting the final point.

    **Expected outcome:** The line does not blink or flash when the last connection is made.

- Start a new A/B Trails activity. Draw a line to an incorrect node.

    **Expected outcome:** The red error line appears without any blink. The error clears after 1 second without any flash.

- Repeat both steps on iOS.

    **Expected outcome:** Behaviour unchanged on iOS.

### ✏️ Notes

- `hideSketchCanvas()` and `useSketchCanvasTimedVisibility` are intentionally kept - they now control the `enabled` flag instead of the mount state, so the 50ms re-enable delay still applies after a wrong connection.
- The fix is isolated to `AbCanvas` and the shared `SketchCanvas` + `useDrawingGesture` primitives. `AbShapes` is untouched.

### DIFF
--- a/src/entities/abTrail/ui/AbCanvas.tsx
+++ b/src/entities/abTrail/ui/AbCanvas.tsx
@@ -78,6 +78,11 @@ export const AbCanvas: FC<Props> = props => {
 
     const id = setTimeout(() => {
       setErrorPath(null);
+      // Clear the live stroke with the error path; guard against a new stroke
+      // started during the error window.
+      if (!getCurrentPath()) {
+        sketchCanvasRef.current?.clear();
+      }
     }, ErrorLineTimeout);
 
     return () => {
@@ -344,7 +349,9 @@ export const AbCanvas: FC<Props> = props => {
       markLastLogPoints({ valid: true });
       addOverCorrectPointToStream(point, time);
       keepPathInState();
-      resetCurrentPath();
+      // Don't clear the live stroke — keep it visible under the committed last
+      // segment so there's no frame where the line disappears (blink).
+      currentPathRef.current = null;
       onMessage(MessageType.Completed);
       incrementCurrentIndex();
       onResult();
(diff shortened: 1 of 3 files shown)


src/entities/abTrail/ui/AbCanvas.tsx
```diff
@@ -366,6 +373,9 @@ export const AbCanvas: FC<Props> = props => {
       const node = findNodeByPoint(point)!;
       hideSketchCanvas();
       setErrorPath(currentPath);
+      // Null the ref so onTouchEnd doesn't call resetCurrentPath() and clear().
+      // Clearing livePath before errorPath renders causes a one-frame blink.
+      currentPathRef.current = null;
```
adeiji, line 378: Thank you for the documentation explaining the why behind the decision

adeiji (approved):
Looks good @sricharan-varanasi , thank you!

PR #1150, opened 2026-06-18
test: test for validating translation lists against English reference (M2-10836)

sricharan-varanasi:
🔗 [Jira Ticket M2-10836](https://mindlogger.atlassian.net/browse/M2-10836)

### 📝 Description

Some translation files use delimiter-separated lists (e.g. `Sun_Mon_Tue`). When a
language has the wrong number of segments, i18next still resolves the value, so the
fallback never triggers and consumers can crash.

Changes include:

- Add `validateTranslations` to compare each language's list values against the English reference by segment count
- Report malformed lists to Datadog on `setupLocalization`, so a bad translation file is caught even after release
- Add a test asserting no list values diverge from the English reference

### DIFF
--- a/src/entities/localization/lib/setupLocalization.ts
+++ b/src/entities/localization/lib/setupLocalization.ts
@@ -3,6 +3,7 @@ import { initReactI18next } from 'react-i18next';
 import { z } from 'zod';
 import { zodI18nMap } from 'zod-i18n-map';
 
+import { getDefaultLogger } from '@app/shared/lib/services/loggerInstance';
 import af from '@assets/translations/af.json';
 import el from '@assets/translations/el.json';
 import en from '@assets/translations/en.json';
@@ -13,6 +14,7 @@ import xh from '@assets/translations/xh.json';
 import zu from '@assets/translations/zu.json';
 
 import { getDefaultLocalizationStorage } from './localizationStorageInstance';
+import { validateTranslations } from './validateTranslations';
 
 declare module 'i18next' {
   interface CustomTypeOptions {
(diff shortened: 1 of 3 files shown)


src/entities/localization/lib/validateTranslations.test.ts
```diff
@@ -0,0 +1,26 @@
...
+describe('translations', () => {
+  // Malformed list values still resolve in i18next, so the fallback never triggers and consumers crash.
+  it('has no list values that diverge from the English reference', () => {
+    const malformed = validateTranslations(
+      resources,
+      REFERENCE_LANGUAGE,
+    ).filter(finding => finding.type === 'malformed-list');
+
+    expect(malformed).toEqual([]);
+  });
```
divbzero, line 25: Should we add a test for `finding.type === 'missing-key'` or is that already tested elsewhere?

src/entities/localization/lib/setupLocalization.ts
```diff
@@ -22,23 +24,40 @@ declare module 'i18next' {
...
+function reportMalformedTranslations() {
+  const malformed = validateTranslations(resources, REFERENCE_LANGUAGE).filter(
+    finding => finding.type === 'malformed-list',
+  );
+
+  malformed.forEach(finding => {
+    getDefaultLogger().warn('[i18n] Malformed translation list', finding);
+  });
+}
+
```
divbzero: Is it necessary to validate translations at runtime when users start the Curious app? Unit tests checking at compile time seems sufficient.

divbzero (approved):
@sricharan-varanasi Looks good overall. I’ve left two comments to consider before merging. Thank you!

src/entities/localization/lib/validateTranslations.test.ts
sricharan-varanasi: added this in a7e5608fa48e18e2962638bd36abc815ecd7e424, thank you

src/entities/localization/lib/setupLocalization.ts
sricharan-varanasi: That makes sense chris, thank you. Addressed this in a7e5608fa48e18e2962638bd36abc815ecd7e424

PR #1152, opened 2026-06-23
fix: AB Trails Android flicker at rendering layer (M2-10920,M2-10921,M2-10922)

sricharan-varanasi:

### 📝 Description

🔗 [Jira Ticket M2-10920](https://mindlogger.atlassian.net/browse/M2-10920)
🔗 [Jira Ticket M2-10921](https://mindlogger.atlassian.net/browse/M2-10921)
🔗 [Jira Ticket M2-10922](https://mindlogger.atlassian.net/browse/M2-10922)

Fixes the Android flicker in AB Trails introduced by the Skia 2.0.2 → 2.4.x upgrade. The upgrade made Android's graphics surface slower to reinitialise, causing a white flash on every segment commit or error. The previous fix resolved the flash but broke gesture reset behaviour, causing free-drawing, multitouch and a stale start point error on iOS.

Changes include:

- Split `SketchCanvas` into a persistent canvas layer and a conditionally mounted gesture layer (`DrawingLayer`)
- Skia `Canvas` stays always mounted - no graphics surface teardown, no flash
- `GestureDetector` and all gesture state still unmount/remount on every hide/show cycle - preserving the original reset behaviour
- Removed `enabled` prop from `useDrawingGesture` - gating is handled by mount/unmount, not flag toggling

### 🪤 Peer Testing

- **Start an AB Trails activity and complete a segment**

    **Expected outcome:** No white flash/blink between segments on Android

- **Draw a line between points without touching them**

    **Expected outcome:** Free drawing is not possible

- **Draw with multiple fingers simultaneously**

    **Expected outcome:** Multitouch is ignored, only one finger tracked

- **Tap an incorrect node, wait for the error to dismiss, then tap the correct node (iOS)**

    **Expected outcome:** Correct node is accepted, no false "Incorrect Start Point" error

### ✏️ Notes

Supersedes the fix on `fix/abtrails-android-flicker`.

### DIFF
--- a/src/shared/ui/SketchCanvas/DrawingLayer.tsx
+++ b/src/shared/ui/SketchCanvas/DrawingLayer.tsx
@@ -1,0 +1,115 @@
+import { FC } from 'react';
+import { StyleSheet, View } from 'react-native';
+
+import { SkPath } from '@shopify/react-native-skia';
+import { GestureDetector } from 'react-native-gesture-handler';
+import { runOnJS, SharedValue, useSharedValue } from 'react-native-reanimated';
+
+import {
+  createLine,
+  getCurrentShape,
+  Point,
+  progressLine,
+  Shape,
+} from './LineSketcher';
+import { useDrawingGesture } from './useDrawingGesture';
+
+type Props = {
+  fullPath: SharedValue<SkPath>;
+  width: SharedValue<number>;
+  onStrokeStart: (x: number, y: number, time: number) => void;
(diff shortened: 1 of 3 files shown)


divbzero (approved):
Thank you @sricharan-varanasi splitting the canvas into a persistent layer makes sense. We should be good to merge!

divbzero:
Looks like Natalia already verified in Android (M2-10921, M2-10922) via APK and will verify in iOS (M2-10920) after merge.

PR #1155, opened 2026-06-29
Updated React Native to 0.85.3

adeiji:
### 📝 Description
Updated RN from  0.79 to 0.85.3. Used rnx-kit/align-deps to update all dependencies necessary to work with RN 0.85.3

 [Jira Ticket M2-#10887](https://mindlogger.atlassian.net/browse/M2-10887)
 [Jira Ticket M2-#10958](https://mindlogger.atlassian.net/browse/M2-10958)
 [Jira Ticket M2-#10967](https://mindlogger.atlassian.net/browse/M2-10967)
 [Jira Ticket M2-#10955](https://mindlogger.atlassian.net/browse/M2-10955)

Changes include:

- **Removed obsolete patches**:
    - patches/react-native+0.79.7.patch
    - patches/react-native-reanimated+3.17.5.patch

- **Added patch**
    - react-native-unity-ios-fabric-init.patch - Under Fabric (new arch), updateProps is not reliably called on RNUnityView, so Unity never initializes. The patch adds [self initUnityModule] to the new arch initWithFrame to match old arch behavior.

- Update the compileSdkVersion for Android to 36
    - Androidx.core:core:1.17.0 is a transitive dependency that was pulled in by one of the libraries upgraded alongside the RN update and it requires compileSdkVersion of 36
- Replaced **react-native-audio-recorder-player** with [**react-native-nitro-sound**](https://www.npmjs.com/package/react-native-nitro-sound). **react-native-audio-recorder-player** is deprecated.
- Replaced @notifee/react-native with react-native-notify-kit
- Updated React to 19.2.3
- Updated dependencies to compatible with 0.85.3
- Set newArchEnabled to true
    - Setting `newArchEnabled=false` in your `gradle.properties` file is not supported anymore since React Native 0.82
- Updated jest.config.ts
- Updated jest.vendor-mocks.js
    - The reanimated v3 Jest setup file (jestUtils.ts) was removed in v4, so we had to add its functionality directly to the mock (default export with components, setUpTests, LinearTransition). The react-native-nitro-sound mocks were consolidated to match the new library's createSound API.
- Removed multiple now obsolete @ts-expect-error directives  
- Added __mocks__ folder
    - react-native-config now depends on native code being available through TurboModules, but Jest only runs JavaScript, so we must manually mock react-native-config


package.json
```diff
@@ -142,24 +139,27 @@
...
+    "react-native-svg": "^15.15.3",
     "react-native-system-navigation-bar": "^2.6.4",
     "react-native-tcp-socket": "^6.0.6",
     "react-native-toast-message": "^2.3.3",
-    "react-native-video": "^6.14.1",
+    "react-native-video": "^6.19.2",
     "react-native-video-player": "^0.16.3",
-    "react-native-webview": "^13.13.1",
+    "react-native-webview": "^13.14.1",
+    "react-native-worklets": "0.11.3",
```
sricharan-varanasi, line 162: A question here, Should reanimated move up with this?
reanimated 4.4.1 only works with worklets 0.9.x, and 0.11.3 is meant for reanimated 4.5.3. Since yarn.lock pins reanimated at 4.4.1, yarn keeps it there and only picks up the new worklets version. Tried it locally and got 4.4.1 + 0.11.3 with an incorrect peer dependency warning. Reanimated only checks this in debug builds, so a release build wouldn't complain.

Would putting worklets back to 0.9.1 be simpler? That's what yarn.lock and Podfile.lock already have. Otherwise we'd bump reanimated to 4.5.3 and re-run pod install.

patches/@azesmway+react-native-unity+1.1.1.patch
```diff
@@ -0,0 +1,54 @@
+diff --git a/node_modules/@azesmway/react-native-unity/ios/RNUnityView.mm b/node_modules/@azesmway/react-native-unity/ios/RNUnityView.mm
+index 6406f64..3c02368 100644
```
sricharan-varanasi, line 2: Does this need a version bump in package.json to go with it?

The filename says 1.1.1, but package.json is still on ^1.0.11 and yarn.lock resolves 1.0.11 - patch-package matches on version, so I don't think this applies.

should we bump the dep to 1.1.1 and regenerate yarn.lock / Podfile.lock?

android/build.gradle
```diff
@@ -2,7 +2,7 @@ buildscript {
     ext {
         buildToolsVersion = "35.0.0"
         minSdkVersion = 25
-        compileSdkVersion = 35
+        compileSdkVersion = 36
```
sricharan-varanasi, line 5: Should AGP move up with this?

Google's [AGP compatibility table](https://developer.android.com/build/releases/about-agp) lists AGP 8.9.1 as the minimum for API 36, but we're pinned to 8.7.3.

patches/@azesmway+react-native-unity+1.1.1.patch
adeiji: Good catch, done.

android/build.gradle
adeiji: Great catch! The update has been made

package.json
adeiji: Change made, thank you @sricharan-varanasi !

package.json
```diff
@@ -142,24 +139,27 @@
     "react-native-markdown-display": "^7.0.2",
     "react-native-mime-types": "^2.4.0",
     "react-native-mmkv": "4.3.1",
-    "react-native-nitro-modules": "0.35.6",
     "react-native-modal-datetime-picker": "^17.1.0",
+    "react-native-nitro-modules": "0.35.6",
+    "react-native-nitro-sound": "^0.2.15",
```
divbzero, line 145: Beyond the React Native upgrade, the migration from `react-native-audio-recorder-player` to `react-native-nitro-sound` seems to be the biggest shift. Have we tested to confirm that audio is working smoothly?

divbzero (approved):
@adeiji Thanks for undertaking this big update. The main question I had was whether we’ve confirmed that audio is working smoothly. Other than that, no major concerns. Thank you!

package.json
adeiji: I left Natalia a note, so she should be testing this.

divbzero:
Rebased against `dev` to resolve merge conflicts in:
- `package.json`
- `yarn.lock`

PR #1157, opened 2026-07-10
fix: Wait for Reset reply from Unity before advancing (M2-10980)

divbzero:
### 📝 Description

🔗 [Jira Ticket M2-10980](https://mindlogger.atlassian.net/browse/M2-10980)

Changes include:

- Wait for Unity to reply to `Reset` message with `"Loaded Successfully"` before advancing
- If Unity is idle in its loading scene (not booting fresh) then probe with `Echo` and proceed on reply

### ✏️ Notes

These changes should work with the Unity code as is. No changes needed in [Unity-Cognitive-Task-Creator](https://github.com/ChildMindInstitute/Unity-Cognitive-Task-Creator).


divbzero:
Superseded by fixes applied by @adeiji in #1155.

PR #1164, opened 2026-08-01
fix: eliminate nav bar jank after unity activity, update touches library, Unity fixes for new architecture (M2-11075, M2-11076, M2-10985, M2-10980)

adeiji:
### 📝 Description
Fixes the layout jank when entering/leaving an activity. The status bar hiding caused the banners padding, screen content, and bottom tab bar to visibly shift.

🔗 [Jira Ticket M2-10985](https://mindlogger.atlassian.net/browse/M2-10985)
🔗 [Jira Ticket M2-11075](https://mindlogger.atlassian.net/browse/M2-11075)
🔗 [Jira Ticket M2-11076](https://mindlogger.atlassian.net/browse/M2-11076)
🔗 [Jira Ticket M2-10980](https://mindlogger.atlassian.net/browse/M2-10980)

Changes include:

After the RN upgrade, Fabric and the newer react-native=screens transition timing caused the native inset resize to become visible mid-transition.

- Enabled edge-to-edge on all Android versions: 
- Migrated 'MainApplication.kt' to RN 0.85's 'loadReactNative() ' entry point (the flat was not working with the old init). Android <= 14 now behaves like 15+, hiding the status bar does not resize the entire window but changes the insets only.
- Applied the stable inset to the three consumers that were shifting: Banners top padding, InProgressActivityScreen margin workaround, and ActivityStepper timer margin.
- Added imperative status bar for Android-only, so the bar fades back during the back transition instead of popping in after. Gated to Android because on iOS the mid-transition restore caused UIKit to recalculate the next screen's header position and size (re-lay), causing a visible jump or shift.
- Unity step transitions: Banners' top padding and screen margin animate to 0 over 250ms when Unity takes over, instead of the previous instant unmount/jump. Both animate back on Unity exit.

## Unity fixes

- **Android:** Unity 6 can't re-initialize after unload in the same process, so the engine now stays alive, parked hidden and paused on unmount, reattached and resumed on remount, with a Reset + app-driven load handshake (the Unity app doesn't announce readiness after a reload).
- **iOS:** opted the Unity view out of Fabric recycling and fixed pause/resume across view instances (frozen/black screen on remount).
- Fixed a stuck fullscreen spinner: a single missed heartbeat during a scene load showed the "unresponsive" overlay and it never cleared on recovery.

## App-wide touch handling
- Migrated touchables (shared TouchableOpacity, moti Pressable via patch, tab bar buttons, stepper buttons) to `react-native-gesture-handler`, and wrapped modals in `GestureHandlerRootView`, fixing intermittently unresponsive taps under the New Architecture.

### 📸 Screenshots

Before
https://github.com/user-attachments/assets/7a6544b3-4eae-4653-a094-4764be3d7dec 

After
https://github.com/user-attachments/assets/32226894-e852-42f7-9863-1b12ad60cf07 

### 🪤 Peer Testing

- [ ] Android 14: open an activity, close it: no status bar jerk, no tab bar resize, no layout shift
- [ ] Android 15+: same as above
- [ ] iOS (notched device): open/close an activity - no content shift at top on return
- [ ] Unity activity: entering/exiting shows smooth top collapse/expand
- [ ] Unity activity: landscape rotation shows no unexpected top gap

### ✏️ Notes

- Tab bar renders behind Android navigation bar after exiting Unity (pre-existing issue, out of scope)


src/entities/banner/ui/Banners.tsx
```diff
@@ -40,7 +42,36 @@ export const Banners = () => {
...
-  const { top } = useSafeAreaInsets();
+  // Use a stable inset so the banners strip keeps reserving the status bar
+  // space while the status bar is hidden during an activity. This prevents
+  // the whole app layout from shifting when the status bar hides/shows.
+  const top = useStableTopInset();
+
+  // Animate the strip's height to 0 when Unity takes over the full screen
+  // (isHidden), instead of unmounting and causing a layout jump. The strip
+  // stays mounted so re-expanding on Unity exit is also smooth.
+  const animatedHeight = useSharedValue(isHidden ? 0 : top);
```
sricharan-varanasi, line 53: Can we use `withTiming` inline here, similar to how `backgroundColor` is handled below?

Something like:

```ts
const animatedStyles = useAnimatedStyle(() => ({
  backgroundColor: withTiming(bannersBg, { ... }),
  paddingTop: withTiming(isHidden ? 0 : top, TIMING),
  marginBottom: withTiming(isHidden ? 0 : insetCorrection, TIMING),
}));
```

It might keep the target value in one place instead of having it both as the shared value’s initial value and again in the effect. That could also avoid a possible first-frame mismatch if those values ever drift.

If this works well here, I think the same approach could potentially be used in `InProgressActivityScreen.tsx:78-85` too.

src/screens/ui/InProgressActivityScreen.tsx
```diff
@@ -48,14 +65,36 @@ export const InProgressActivityScreen: FC<Props> = ({ navigation, route }) => {
...
+
+  useEffect(() => {
+    animatedMargin.value = withTiming(targetMargin, {
+      duration: 250,
+      easing: Easing.out(Easing.ease),
+    });
+  }, [targetMargin, animatedMargin]);
+
+  const animatedContainerStyle = useAnimatedStyle(() => ({
+    flex: 1,
```
sricharan-varanasi, line 88: nit: as flex: 1 is static, so it gets recomputed on the UI thread every frame here. Could we move it into a StyleSheet and compose, style={[styles.container, animatedContainerStyle]}?

src/features/pass-survey/ui/ActivityStepper.tsx
```diff
@@ -59,7 +57,7 @@ export function ActivityStepper({
   onSkipActivity,
 }: Props) {
   const { t } = useTranslation();
-  const { top } = useSafeAreaInsets();
+  const top = useStableTopInset();
```
sricharan-varanasi, line 60: minor: Could this end up capturing different inset values depending on whether the applet is cached? On a first load, the stepper may mount after the status bar is already hidden and get `top = 0`, while a cached load could still get the actual inset.

Since line 353 depends on `top === 0`, that could cause a small layout difference between the two cases.

Would it make sense for the hook to cache the inset once app-wide instead? `Banners` mounts early enough to capture the correct value, and later components could reuse it.

src/entities/banner/ui/Banners.tsx
adeiji: @sricharan-varanasi Thanks!  Good catch; changes have been made.

src/entities/unity/lib/hook/useUnityLifecycle.ts
```diff
@@ -248,11 +419,11 @@ export const useUnityLifecycle = (options: UseUnityLifecycleOptions) => {
   );
   useEffect(() => {
     registerEventHandler(UnityEventSetOrientation, handleSetOrientation);
-    return () => {
-      RNOrientationDirector.lockTo(Orientation.portrait);
-    };
```
divbzero, line 253: Did we intend to remove `lockTo(Orientation.portrait)` on unmount? Seems like even with `unlock()` on mount, we should keep `locate(Orientation.portrait)` on unmount.

package.json
```diff
@@ -140,14 +140,15 @@
...
-    "react-native-nitro-modules": "0.35.6",
+    "react-native-nitro-modules": "^0.37.1",
     "react-native-nitro-sound": "^0.2.15",
     "react-native-notify-kit": "^10.4.6",
     "react-native-orientation-director": "^2.6.5",
     "react-native-permissions": "^4.0.1",
     "react-native-popover-view": "^6.0.1",
     "react-native-progress": "^5.0.1",
-    "react-native-reanimated": "^4.4.0",
+    "react-native-quick-base64": "^3.0.1",
```
divbzero: Is `react-native-quick-base64` being used? I can’t find a reference in the code.

divbzero (approved):
@adeiji This looks pretty solid. Beyond resolving merge conflicts, only two places where I noticed something to double check. Good to merge afterwards. Thank you!

src/entities/unity/lib/hook/useUnityLifecycle.ts
adeiji: Yeah, you're right. I think now that we've diagnosed what the issue was with Unity, this can be brought back.

package.json
adeiji: Good catch, that was there from the encryption library change but it's unecessary now.

sricharan-varanasi (approved):

divbzero:
Rebased against `fix/rn-update-m2-10887` to resolve conflicts in:
- `package.json`
- `yarn.lock`
- `ios/Podfile.lock`
- `patches/@azesmway+react-native-unity+1.1.1.patch`
- `src/entities/applet/lib/hooks/useAppletAnalytics.ts`

PR #1165, opened 2026-08-06
Random MMKV Encrypt Key

adeiji:
- [x] Related documentation has been added/updated

### 📝 Description

Replaces the build-time MMKV encryption key (`STORE_ENCRYPTION_KEY` from `.env`, which was identical for every install and at build time was embedded in the binary) with a random per-device key generated at first launch and stored in iOS Keychain / Android Keystore using react-native-keychain. Existing users are migrated by recrypting all secured stores using the legacy STORE_ENCRYPTION_KEY.

🔗 [Jira Ticket M2-10992](https://mindlogger.atlassian.net/browse/M2-10992)

Changes include:

- At launch in mmkvEncryptionKeyManager.ts, we resolve the key: read it from the keychain/store or generate a 32-hex random key and persist it before migrating (recrypting).  Then we recrypt all encrypted stores with our new random key.
- Recrypted stores are flagged as migrated in the plain `system` store after each store finishes recryption. A store that fails to recrypt (e.g. very rare, but a corrupted store) is wiped and logged to Datadog for our reference.
- The StorageEncryptionProvider acts as a gate ensuring that the encryption key has been resolved before any part of the app can touch encrypted stores.
- Both BackgroundWorker handlers (the in-app fetch callback and the Android headless task - app killed) await key resolution before running. If the device has not yet been unlocked after a reboot, we skip the run and let the next scheduled fetch retry.
- In the very rare case that the device has a broken keystore, we fall back to the legacy env key with a Datadog error and retry again later.  This fallback is set just in case this happens; we'll have logged it and be able to see which devices are having this issue, if any.

### 🪤 Peer Testing

  Requires `yarn install` + `yarn pods`

**Upgrade path:**
1. Check out `dev`, build & run on device/sim
2. Log in, start an activity, stop partway, close the app
3. Check out this branch, `yarn install` + `yarn pods`
4. Rebuild & run — do NOT uninstall the app first
5. Confirm: still logged in, activity resumes where you left off

**Fresh install:**
1. Uninstall the app
2. Build & run this branch
3. Log in, kill the app, relaunch
4. Confirm: still logged in

### ✏️ Notes

- The Legacy key is retained for now for the duration of the migration window. This needs to be removed once a sufficient percentage of devices has migrated.
- Failures in background tasks are now logged to Datadog so we know if our changes cause issues in production.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,6 +1,8 @@
 ENV=dev
 API_URL=http://localhost:8000
 MIXPANEL_TOKEN=
+# Legacy MMKV encryption key, only used to migrate old installs to the random
+# per-device key (and as a fallback when the hardware keystore is unavailable)
 STORE_ENCRYPTION_KEY=my-encryption-key
 LAUNCHDARKLY_MOBILE_KEY=my-env-mobile-key
 ONEUP_HEALTH_CLIENT_ID=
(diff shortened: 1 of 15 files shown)


divbzero (approved):
@adeiji Changes seem reasonable to me, thank you. Let’s wait for QA confirmation before merging.

PR #1170, opened 2026-09-04
fix: Override `Event.canCoalesce()` to return `false` (M2-11083)

divbzero:
### 📝 Description

🔗 [Jira Ticket M2-11083](https://mindlogger.atlassian.net/browse/M2-11083)

React Native defaults `Event.canCoalesce()` to return `true` which drops earlier events of the same type when multiple are queued. This surfaced in the Android app as “Unity Error” and ”Unity Build Failed” popups.

Changes include:
- Override `Event.canCoalesce()` to return `false`.

### ✏️ Notes

Same fix as wonday/react-native-pdf#1011.

### Functionality

### Testing

- [ ] Test suite passes

### Security & Data Privacy

### Logging/Monitoring

### Performance

### Readability

### Change Safety


src/shared/ui/BackButton.tsx
```diff
@@ -1,21 +1,22 @@
 import { PropsWithChildren } from 'react';
-import { TouchableOpacity } from '@app/shared/ui/TouchableOpacity';
 
 import { useNavigation } from '@react-navigation/native';
 
+import { TouchableOpacity } from '@app/shared/ui/TouchableOpacity';
```
adeiji: I fixed this in #1164 after you created your fix branch. Do you want to leave it as is and handle it when we merge, or should we take it out now?

adeiji (approved):
Looks good @divbzero !

divbzero:
Rebased against `M2-11079-react-navigation-header-inset` to resolve conflicts in:
- `package.json`
- `yarn.lock`
- `ios/Podfile.lock`
- `patches/@azesmway+react-native-unity+1.1.1.patch`
- `src/entities/activity/ui/ActivityCard.tsx`
- `src/entities/banner/ui/Banners.tsx`
- `src/entities/unity/lib/hook/useUnityLifecycle.ts`
- `src/screens/ui/InProgressActivityScreen.tsx`
- `src/shared/lib/hooks/useStableTopInset.test.ts`
- `src/shared/lib/hooks/useStableTopInset.ts`

PR #1172, opened 2026-09-08
fix: added rounded border to activity card and less opacity when disa…

adeiji:
…bled

### 📝 Description

The corners of the activity cards lost their rounded corners during the RN upgrade. This PR adds the rounded corner styling back, as well as the .5 opacity when the activity card is disabled.

🔗 [Jira Ticket M2-#11117](https://mindlogger.atlassian.net/browse/M2-11117)

### 📸 Screenshots

Before
card corner issue
------------------------- | ------------------------------------- |

After                                 
Screenshot 2026-09-08 at 11 00 21 AM

### Functionality

### Testing

- [ ] Test suite passes

### Security & Data Privacy

### Logging/Monitoring

### Performance

### Readability

### Change Safety


src/entities/activity/ui/ActivityCard.tsx
```diff
@@ -57,6 +57,10 @@ export const ActivityCard: FC<Props> = ({
       aria-label={accessibilityLabel}
       onPress={onPress}
       disabled={isDisabled}
+      style={{
+        borderRadius: 16,
+        opacity: disabled ? 0.5 : 1,
```
divbzero: Should we be using `isDisabled` here as well?

src/entities/activity/ui/ActivityCard.tsx
divbzero, line 66: How does `<YStack opacity={isDisabled ? 0.5 : 1}>` interact with `<AnimatedTouchable style={{ opacity: disabled ? 0.5 : 1 }}>` above? Maybe we just need one of these?

divbzero (approved):
@adeiji Maybe one quick fix on `opacity` but otherwise looks good!

src/entities/activity/ui/ActivityCard.tsx
adeiji: Good catch, done.

src/entities/activity/ui/ActivityCard.tsx
adeiji: Removed according to comment below.

divbzero:
Rebased against `fix/activity-status-bar-jank-m2-10985` to resolve conflicts in:
- `package.json`
- `yarn.lock`
- `ios/Podfile.lock`
- `src/entities/activity/ui/ActivityCard.tsx`
- `src/entities/unity/lib/hook/useUnityLifecycle.ts`

PR #1177, opened 2026-09-11
fix: Hidden email input on Android `SignUpScreen` (M2-11126, M2-11146)

divbzero:
### 📝 Description

🔗 [Jira Ticket M2-11126](https://mindlogger.atlassian.net/browse/M2-11126)
🔗 [Jira Ticket M2-11146](https://mindlogger.atlassian.net/browse/M2-11146)

On Android, we use [`android:windowSoftInputMode="adjustPan"`](https://developer.android.com/guide/topics/manifest/activity-element.html#wsoft) to pan the window and keep input fields visible when the keyboard is displayed. However, we also wrap `SignUpScreen` with [`KeyboardAvoidingView`](https://reactnative.dev/docs/keyboardavoidingview) which shifts input fields up when the keyboard is displayed. The use of both `adjustPan` and `KeyboardAvoidingView` resulted in the email input field on `SignUpScreen` to be pushed up too much and be hidden by the header.

Changes include:

- Switch `KeyboardAvoidingView` from `behavior="position"` to `behavior="padding"`. This reduces padding instead of shifting the position of the view.
- Scroll password input to top instead of scroll view to end, which was hiding the password input on smaller screens.
- Measure `footerHeight` instead of hardcoding, which was causing the footer to be partially visible on smaller screens.

### Functionality

### Testing

- [ ] Test suite passes

### Security & Data Privacy

### Logging/Monitoring

### Performance

### Readability

### Change Safety


sricharan-varanasi (approved):
Looks good, @divbzero

PR #1178, opened 2026-09-11
fix: limit heartbeat tick logging to non-prod

sricharan-varanasi:
### 📝 Description

🔗 [Jira Ticket M2-11085](https://mindlogger.atlassian.net/browse/M2-11085)

Changes include:

- Log heartbeat ticks only in dev and UAT.
- Do not log heartbeat ticks in production.
- Keep heartbeat failure logging enabled in all environments.

### 🪤 Peer Testing

- Run the Unity heartbeat flow in dev and UAT.  
  **Expected outcome:** Heartbeat ticks are logged.

- Run the Unity heartbeat flow in production.  
  **Expected outcome:** Heartbeat ticks are not logged; failures remain logged.

- Automated test: `useUnityHeartbeat.test.ts`  
  **Result:** 6 tests passed.

### Functionality

- [x] The feature behaves correctly in practice and fulfills the intended business purpose
- [x] The implementation accounts for edge cases, avoids subtle logical errors, and handles somewhat rare failure states (e.g. offline mode for mobile, 3rd party being down, etc)

### Testing

- [x] Verify there are automated tests added that meaningfully cover critical behavior and failure cases
- [x] Code coverage does not go down as result of this change
- [x] Test suite passes

### Security & Data Privacy

- [x] Verify there is no chance we would accidentally log PII to application logs
- [x] Verify this addition does not materially affect our security attack surface, and if so it has undergone security review
- [x] All inputs are sanitized
- [x] New dependencies are well maintained, have significant justification for being added to the project, and are documented in the Curious [open source credit page](https://mindlogger.atlassian.net/jira/servicedesk/projects/MLA/knowledge/articles/340623543?spaceKey=MLA)

### Logging/Monitoring

- [x] Logging is implemented for this change such that you could troubleshoot this feature in production
- [x] The change/feature is able to be monitored in production

### Performance

- [x] This change does not introduce n+1 queries or other performance issues within our expected scale (e.g. missing indexes on frequently queried columns, frequently updating tables that are accessed often)

### Readability

- [x] All commented out code is removed
- [x] Debugging code including extraneous log lines are removed
- [x] Code is easy to understand through naming and structure; comments explain intent or non-obvious decisions

### Change Safety

- [x] ~~Destructive database migrations are rolled out in stages. For example, renaming a column means adding a new column and migrating existing data to that columns in one deployment. Then monitoring to ensure that field isn’t used, and finally removing that old field in a separate deployment.~~


divbzero (approved):
Looks good, thanks @sricharan-varanasi.

src/entities/unity/lib/hook/useUnityHeartbeat.ts
divbzero, line 74: @sricharan-varanasi Actually, I think we should also update `sendMessageToUnity` to avoid logging `Echo` messages in production.

divbzero (changes requested):

src/entities/unity/lib/hook/useUnityHeartbeat.ts
divbzero: Fixed in commit ab2e919. Thanks @sricharan-varanasi.

src/entities/unity/lib/hook/useUnityHeartbeat.ts
```diff
@@ -62,7 +63,9 @@ export const useUnityHeartbeat = ({
     intervalRef.current = setInterval(() => {
       const echoPayload = `heartbeat-${Date.now()}`;
       const echoMsg = newEchoMessage(echoPayload);
-      logger.log('[Heartbeat] sending Echo tick');
+      if (ENV === 'dev' || ENV === 'uat') {
```
divbzero, line 66: @sricharan-varanasi For consistency with the how we filter the echo message, we should probably change this from `ENV === 'dev' || ENV === 'uat'` to `ENV !== 'production'`.

divbzero (approved):
Thanks @sricharan-varanasi. Approved, but maybe update the `ENV` condition to be consistent before merging.

PR #1180, opened 2026-09-11
fix: Fix flanker activity crashing on iOS

adeiji:
### 📝 Description

On the old architecture there was only one FlankerViewManager, so caching the view on the manager instance worked. On the new architecture there are multiple manager instances (the one that creates the view is not the one that receives NativeModules.startGame), so its cached view was always nil and the force-unwrap crashed.

The view now registers itself in a static weak reference (FlankerView.current) that all manager instances can read. startGame resolves the view from there, asserts in dev builds if no view is live (no-ops in release instead of crashing).

🔗 [Jira Ticket M2-#11136](https://mindlogger.atlassian.net/browse/M2-11136)

### Functionality

### Testing

- [ ] Test suite passes

### Security & Data Privacy

### Logging/Monitoring

### Performance

### Readability

### Change Safety


divbzero:
Rebased against `M2-11131-react-native-reanimated-4.4.2`. No conflicts.

divbzero (approved):
The fix makes sense and the code looks good, thanks @adeiji!

PR #1186, opened 2026-09-17
fix: Prevent applet card gaps after Unity landscape rotation on Android (M2-11130)

adeiji:
### 📝 Description

🔗 [Jira Ticket M2-11130](https://mindlogger.atlassian.net/browse/M2-11130)

After completing a Unity activity on Android, returning to the applet list showed large gaps between tiles. Unity rotates the app to landscape and back, and the resulting width change leaves the `FlatList`'s native cells with out-of-sync offsets.

The list is now keyed on window width on Android, which rebuilds the cells and offsets after a rotation. The key deliberately ignores height-only changes (system bars showing/hiding) so the list isn't reset unnecessarily.

Changes include:

- Key the applet `FlatList` on window width (Android only) so it rebuilds after Unity's landscape rotation

### 🪤 Peer Testing

- Open the applet list and note the tile spacing
**Expected outcome:** Tiles are evenly spaced

- Start and complete a **Unity activity**, then return to the applet list
**Expected outcome:** Tiles are still evenly spaced, with no large gaps

- Scroll the list and tap a tile below the fold
**Expected outcome:** The list scrolls normally, and the tile opens

### Functionality

### Testing

- [ ] Test suite passes

### Security & Data Privacy

### Logging/Monitoring

### Performance

### Readability

### Change Safety


divbzero (approved):
@adeiji Interesting fix, makes sense to me 👍

PR #1187, opened 2026-09-17
fix: Remove `opacity: 0` from CST control bar mask so it stays tappable (M2-11141)

adeiji:
### 📝 Description

🔗 [Jira Ticket M2-11141](https://mindlogger.atlassian.net/browse/M2-11141)

The "Tap here to start" control for CST activities did not respond to taps. The control captures touches through an invisible full-size mask that was hidden with `opacity: 0`. The old architecture could still receive touches on views with opacity: 0. With the New Architecture, native hit-testing ignores fully transparent views, so the mask no longer received touch events.

The mask has no background or children, so it is invisible without the opacity; removing it restores tap handling with no visual change.

Changes include:

- Remove `opacity: 0` from the CST control bar mask

### 🪤 Peer Testing

- Start a **CST activity in gyroscope mode** and tap "Tap here to start"
**Expected outcome:** The test begins

- Start a **CST activity in touch mode** and tap "Tap here to start"
**Expected outcome:** The test begins

### Functionality

### Testing

- [ ] Test suite passes

### Security & Data Privacy

### Logging/Monitoring

### Performance

### Readability

### Change Safety


divbzero (approved):
Looks good, thanks @adeiji!

PR #1190, opened 2026-09-18
fix: Nav bar extra space in Unity (M2-11155)

adeiji:
### 📝 Description

🔗 [Jira Ticket M2-11155](https://mindlogger.atlassian.net/browse/M2-11155)

After completing a Unity activity on Android, returning to the applet screen showed the nav bar as taller then before.

The cause is a bug in react-native-screens 4.25.2. With `unstable_headerInsets: { top: false }` (added in a0d86908 as a fix for ticket [M2-11079)](https://mindlogger.atlassian.net/browse/M2-11079) a0d86908 screens still added that disabled inset to the measured header height. The bad measurement only surfaced after Unity activity, whose landscape rotation forces a re-layout. Fixed upstream in 4.27.0 by [software-mansion/react-native-screens#4452](https://github.com/software-mansion/reactnative-screens/pull/4452)

Reproduced only on Android, and only on devices without a persistent navigation bar.
                                                                                    
Changes include:
                                                                                    
- Upgrade `react-native-screens` 4.25.2 -> 4.27.0, pinned exactly
                                                                                    
### 🪤 Peer Testing
                                                                                    
**Requires `yarn install`**
                                                                                    
**Android, on a gesture-navigation device (e.g. Pixel). Does not reproduce with 3-button navigation.**                                                                       

- Open an applet and note the tab bar

**Expected outcome:** Tab bar sits flush at the bottom of the screen

- Start and complete a **Unity activity**, then return to the applet screen

**Expected outcome:** Tab bar is the same height as before, with no gap beneath it

- Navigate between the Activities / Data / About tabs

**Expected outcome:** Tab bar stays correct

**iOS regression check** — screens affects every screen transition:

- Navigate through applet details, activities, and settings

**Expected outcome:** No layout or transition changes

### Functionality

### Testing

- [ ] Test suite passes

### Security & Data Privacy

### Logging/Monitoring

### Performance

### Readability

### Change Safety


divbzero:
Rebased against `fix/m2-11141-cst-start-issue` to add to stack of React native upgrade PRs.

divbzero (approved):
Thank you @adeiji!

